Issue #12, opened 2014-02-03
get out of savannah


cbernet:
We are migrating to github for issue tracking. 

We still depend on savannah for the publication of the CMG samples. We decided to stop publish information to savannah, and to rely on the sole CMG DB to keep track of our samples. 

Somebody should take care of removing the corresponding code from the CMGTools/Production package.

gpetruc:
I was aware only of the things that are now fixed in the pull request #16.
I'm leaving this open for another few days if there's anything else that people think we have to do related to savannah outside that package (@cbernet @ssekmen: any ideas?)

PR #16, opened 2014-02-04
Cleanup production tools cmg pat v5 18 from cmssw 5 3 14

gpetruc:
A production tools version that is mostly free of savannah and git, for #12
(to be tested before merging)


cbotta:
Ciao Giovanni,
I've tested all the steps of the production chain with the new version (PFAOD, cmgTuple, resubmission, intregritychecks, publishing, tree production with the python framework)
It is ok.

Thanks,
Cristina

PR #32, opened 2014-03-20
Switch to das_client to get dbs3 files

gpetruc:
Needed for recent datasets that are not visible with dbsql

Would benefit from more testing  (in particular on a dataset with more than 10k files)

Now run ranges are also implemented, but they can be extremely slow and result in timeout errors (after ~5 mins or so), especially if using ranges where one end is trivial (e.g. [1, 203986]).
In principle one could implement a workaround querying first the list of runs and then the files for each run, but that will also be quite slow (e.g. there's ~500 runs in the Run2012D era).


gpetruc:
@cbotta @ssekmen  I have merged this already into the V5_18 working area of cmgtools

PR #41, opened 2014-03-30
Wrappers to aliases in tutorial.

adavidzh:


CMGTools/RootTools/scripts/multiloop
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+ipython -i $CMSSW_BASE/src/CMGTools/RootTools/python/fwlite/MultiLoop.py "$@"
```
gpetruc, line 2: Uhm, maybe by default i'd have it without the <b>-i</b> option? have it with -i can be instructive at the beginning but it's annoying when one is actually testing something that runs up to completion (especially if running with split > 1 so that the outputs all mixes up and one sometimes can't tell if the script is waiting for the final ctrl+D or is still running)

adavidzh, line 2: I see your point. Then perhaps this and `htmldir` both ought to be changed from `ipython -i` to just `python`, but  `pyroot` still makes sense to be `ipython -i`, right?

PR #48, opened 2014-05-12
add various mt2, and Isolated track

mariadalfonso:
Added various mt2 class.
Added also two analyzers: one for IsolatedTrackVeto, the other for all the topological variables like mt, mt2, mt2w

### DIFF
--- a/CMGTools/TTHAnalysis/python/analyzers/ttHIsoTrackAnalyzer.py
+++ b/CMGTools/TTHAnalysis/python/analyzers/ttHIsoTrackAnalyzer.py
@@ -25,7 +25,7 @@ class ttHIsoTrackAnalyzer( Analyzer ):
     #----------------------------------------
     def declareHandles(self):
         super(ttHIsoTrackAnalyzer, self).declareHandles()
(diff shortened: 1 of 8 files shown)


CMGTools/TTHAnalysis/python/analyzers/ttHIsoTrackAnalyzer.py
```diff
@@ -0,0 +1,140 @@
...
+    def __init__(self, cfg_ana, cfg_comp, looperName ):
+        super(ttHIsoTrackAnalyzer,self).__init__(cfg_ana,cfg_comp,looperName)
+
+    #----------------------------------------
+    # DECLARATION OF HANDLES OF LEPTONS STUFF   
+    #----------------------------------------
+    def declareHandles(self):
+        super(ttHIsoTrackAnalyzer, self).declareHandles()
+#        self.handles['cmgCandidates'] = AutoHandle('cmgCandidates','std::vector<cmg::Candidate>')
+        self.handles['cmgCand'] = AutoHandle('cmgCandidates','std::vector<cmg::Candidate>')
```
gpetruc: it would be good to take the label and the type of the cmgCand as parameters in the cfg, since we probably need different values for V5_18, V5_17 and CMGPF V5_16 samples

PR #57, opened 2014-06-27
moved met and HadTop out from the ttHLepEventAnalyzer; added QD discrimi...

mariadalfonso:
...ator tot he jets

### DIFF
--- a/CMGTools/TTHAnalysis/python/analyzers/ttHJetAnalyzer.py
+++ b/CMGTools/TTHAnalysis/python/analyzers/ttHJetAnalyzer.py
@@ -70,13 +70,25 @@ class ttHJetAnalyzer( Analyzer ):
         event.jets = []
         event.jetsFailId = []
         for jet in allJets:
+
+            rho  = float(self.handles['rho'].product()[0])
+            if self.jetReCalibratorCHS:
+                jet.QG = jet.quarkGluonID(rho, "LD_CHS_CMGVARS","Pythia")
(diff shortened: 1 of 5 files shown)


gpetruc:
Hi Maria,

As is, it doesn't work as it conflicts with changes in #54, and since all the changes are included in a single commit I can't directly pick only the changes that are still ok.
I'll close this, and put in another PR the addition of the jet info variables (QG likelihood and pileup jet id)
Do you really want the old ttH combinatorical reconstruction of the ttbar decay in the topological analyzer? in #54 I've moved it to be only in the lepton config. 

Giovanni

CMGTools/TTHAnalysis/python/analyzers/ttHJetAnalyzer.py
```diff
@@ -70,13 +70,25 @@ def process(self, iEvent, event):
         event.jets = []
         event.jetsFailId = []
         for jet in allJets:
+
+            rho  = float(self.handles['rho'].product()[0])
+            if self.jetReCalibratorCHS:
+                jet.QG = jet.quarkGluonID(rho, "LD_CHS_CMGVARS","Pythia")
+                #    if(jet.pt()>50):
+                #        print 'QG: ',jet.quarkGluonID(rho, "LD_CHS_CMGVARS","Pythia")
```
gpetruc, line 78: I don't think data/mc corrections for pythia are supposed to be applied also to data...

PR #75, opened 2014-07-07
added the MET triggers

mstoye:


gpetruc:
Hi Markus,

The PFMET150 was already there, can you change the commit so to add only the parked one?

Thanks

gpetruc:
Hi Markus,

I had probably missed that you had added some more commits, but now I see that the whole thing is  a bit more complicated than just adding a trigger and github doesn't know how to merge it anyway because other things have gone in 53X in the meantime.

At the moment at which you have a stable configuration, can you do the merging or rebasing manually and make a new pull request? 
I leave this open as a reminder

PR #82, opened 2014-07-10
maria_7Z_v2

mariadalfonso:


gpetruc:
Hi Maria,

For the tau in 7.X, the dxy should be available from the method of the pat::Tau:
    https://github.com/cms-sw/cmssw/blob/CMSSW_7_0_X/DataFormats/PatCandidates/interface/Tau.h#L314
for the dz I don't know.

Anyway, will test later tomorrow and merge

PR #83, opened 2014-07-11
added nPU to event

pandolf:


gpetruc:
I suppose you also want to add it in the output tree, right?

Then please also modify **ttHLepTreeProducerNew.py** look whatever it does for **puWeight** and apply it also to **nPU**, then commit and push again so that the pull request is updated.

Thanks

Giovanni

pandolf:
done.

PR #93, opened 2014-07-18
Pr92 remerged

mariadalfonso:


CMGTools/TTHAnalysis/python/analyzers/ntupleTypes.py
```diff
@@ -195,12 +195,14 @@
 ##------------------------------------------  
 
 genParticleType = NTupleObjectType("genParticle", baseObjectTypes = [ particleType ], mcOnly=True, variables = [
-    NTupleVariable("charge",   lambda x : x.threeCharge()/3.0, float),
+    NTupleVariable("charge",   lambda x : x.threeCharge()/3.0, int),
```
gpetruc: There's a reason for this to be a float: genParticles like quarks have non-integer charges.
Please revert this change in a next commit.

PR #94, opened 2014-07-18
this is a backport of the 70X committed in #93

mariadalfonso:


CMGTools/TTHAnalysis/python/analyzers/ttHJetMCMatchAnalyzer.py
```diff
@@ -57,7 +57,7 @@ def isFlavour(x,f):
     def matchJets(self, event):
         match = matchObjectCollection2(event.cleanJetsAll,
                                        event.genbquarks + event.genwzquarks,
-                                       deltaRMax = 0.5)
+                                       deltaRMax = 0.3)
```
gpetruc: In 5.3.X we have AK5 jets, so I'd leave the matching cone as is (or maybe put it to 0.4?)

PR #100, opened 2014-07-24
WMass and hadd changes

perrozzi:
Quite large W  mass commit...
Also added small option to hadd


CMGTools/Production/python/hadd.py
```diff
@@ -77,7 +77,7 @@ def haddRec(odir, idirs):
         for file in files:
             hadd('/'.join([root, file]), odir, idirs)
 
-def haddChunks(idir, removeDestDir, cleanUp=False):
+def haddChunks(idir, removeDestDir, cleanUp=False, outputdir='', odir_cmd='./'):
```
gpetruc: Why do you have two parameters "outputdir" and "odir_cmd" ? (of wihch outputdir seems to be not used)

perrozzi: leftover, now removed

PR #105, opened 2014-08-06
Cleanup unused XML files

gpetruc:
To my knowledge nothing of this is needed anymore in 7.0.X ( @steggema please confirm for the tau tau side )


steggema:
Hi Giovanni,

While we are currently still using the PU jet ID weight files in CMGTools/External for the MVA MET producer as a temporary solution, we should be able to use the ones from RecoJets/JetProducers/, so you can go ahead with the PR.

Cheers,
Jan

PR #110, opened 2014-08-13
Local db sfor private mini aod

mmasciov:
I have added few methods in order to access privately produced miniAOD samples.
The additions involve the module ComponentCreator.py and dataset.py. 
The rest of the code in these modules is unchanged.
I have also added the list of privately produced miniAOD samples with bx=50 ns, in samples_13TeV_CSA14.py.


gpetruc:
Hi,

It seems to me that in the way the code was added the DBS results are not cached (unlike what it happens normally for DBS results retrieved with the standard component builder specifying user="CMS").
This slows down jobs startup time, and would result in tons of DBS queries when running in the batch system, so it's not acceptable.

Please modify things so that the DBS result is cached, as in the other cases.

mmasciov:
Hi,

I added that feature. To do that, I have just created some methods which are parallel to the original ones. They could also be merged, but I preferred to keep them separate for the time being.

Cheers,
Mario

PR #123, opened 2014-09-19
First commit of AlphaT configurations and code

MarkBaber:
Preliminary commit for adding the AlphaT analysis to the CMG framework for the Cmg_mini_aod_lite_v6_0_from_cmssw_7_0_6 branch


gpetruc:
checking. 
@mariadalfonso can you also maybe have a look, since it's hadronic susy?

CMGTools/TTHAnalysis/cfg/run_susyAlphaT_cfg.py
```diff
@@ -0,0 +1,209 @@
...
+ttHJetAna.jetLepDR        = 0.4
+ttHJetMCAna.smearJets     = False
+
+
+# Energy sums
+#------------------------------
+# NOTE: Currently energy sums are calculated with 40 GeV jets (ttHCoreEventAnalyzer.py)
+#       However, the input collection is cleanjets which have a 50 GeV cut so this is a labeling problem
+
+ttHJetMETSkim.htCut       = ('htJet50j', 0)
```
gpetruc, line 64: Here this works only because the cut is at zero, since the variable htJet50j is not defined, though since you raised the threshold on jets to 50 GeV then the contents of the htJet40j will reproduce what htJet50j should be. 
So, if you want to apply a non-zero cut, you should do it on 'htJet40j', or add the htJet50j in the core analyzer.
Anyway I'll merge this pull request as is, since it creates no problems. 
Mariarosaria @mariadalfonso is looking into reorganizing a bit the set of definitions of HTs and MHTs since now it's a mess of many copy-pastes

PR #124, opened 2014-09-28
Hemi control var script

mariadalfonso:
Changes of general interest.
a) submission from T3 other then CERN
b) added option to save uncompressed files
c) gen lepton matching: 
    updated tau vector for P8
    renamed branches GenLep to genLep 
    moved out the GenTop,GenBQuark,GenQuark from the treeProducerSusyCore
d) added option for E/mu cross cleaning
e) added SMS T1bbbb and T1tttt from the minAOD to the list CSA14
f) adopted convention for eventSums variables with jets only and lepton+jets:
 suffix _had added for the final variable with input jets only ( input vector name has a j added), no suffix with the jets and lepton+jets:

MT2 specific:
a) script for merging baby tree
b) Hemisphere: add kt and akt algo, and bug fix to default, added mt2bb
c) trigger for control sample
d) control sample variables
e) renamed run_susyFullHad_cfg.py to run_susyMT2_cfg.py


CMGTools/Production/python/batchmanager.py
```diff
@@ -47,35 +47,50 @@ def DefineOptions(self):
...
-            else:
-                # directory exists.
-                if self.options_.negate is False and self.options_.force is False:
-                    #COLIN need to reimplement protectedRemove in eostools
-                    raise ValueError(  ' '.join(['directory ', self.remoteOutputDir_, ' already exists.']))
-                    # if not castortools.protectedRemove( self.remoteOutputDir_, '.*root'):
-                    # the user does not want to delete the root files                          
+    
+            if "pnfs" in self.remoteOutputDir_: # T3 @ PSI:
+                if self.remoteOutputDir_.startswith("/pnfs"):
```
gpetruc, line 59: Shouldn't the first if be "if self.remoteOutputDir_.startswith("/pnfs")"  (since "pnfs" itself might legitimately appear anywhere in a directory name), and the second if should test if the path is somehow related to T3@PSI ?
Even better, can you put the hostname as a prefix in the path?
Otherwise, what do we do when somebody else wants to write anywere beside PSI ? (currently, besides CERN this framework is being used by IC, LLR, DESY...)

CMGTools/TTHAnalysis/python/samples/samples_13TeV_CSA14.py
```diff
@@ -166,11 +170,31 @@
...
+
+#mcSamplesCSA14_PU40bx50 = [ SMS_T1bbbb_2J_mGl1000_mLSP900_PU_S14_POSTLS170, SMS_T1bbbb_2J_mGl1500_mLSP100_PU_S14_POSTLS170, SMS_T1qqqq_2J_mGl1400_mLSP100_PU_S14_POSTLS170, SMS_T1tttt_2J_mGl1200_mLSP800_PU_S14_POSTLS170, SMS_T1tttt_2J_mGl1500_mLSP100_PU_S14_POSTLS170 ]
+
+
+##T1tttt2J_6_PU_S14_POSTLS170 = kreator.makeMyPrivateMCComponent("/SMS-T1tttt_2J_mGl-1500_mLSP-100_Tune4C_13TeV-madgraph-tauola/phys_susy-miniAODforSusy_SMS-T1tttt_2J_mGl-1500_mLSP-100_Tune4C_13TeV-madgraph-tauola_PU_S14_POSTLS170-af38aa319b7b7c91a6797b31c3be19b7/USER", "PRIVATE", ".*root", "phys03")
+##T1tttt2J_7_PU_S14_POSTLS170 = kreator.makeMyPrivateMCComponent("/SMS-T1tttt_2J_mGl-1200_mLSP-800_Tune4C_13TeV-madgraph-tauola/phys_susy-miniAODforSusy_SMS-T1tttt_2J_mGl-1200_mLSP-800_Tune4C_13TeV-madgraph-tauola_PU_S14_POSTLS170-af38aa319b7b7c91a6797b31c3be19b7/USER", "PRIVATE", ".*root", "phys03")
```
gpetruc, line 195: Why do we have 2 copies of T1tttt + a third one commented out?

CMGTools/Production/python/batchmanager.py
mariadalfonso, line 59: see https://github.com/CERN-PH-CMG/cmg-cmssw/pull/125

CMGTools/TTHAnalysis/python/samples/samples_13TeV_CSA14.py
mariadalfonso, line 195: see https://github.com/CERN-PH-CMG/cmg-cmssw/pull/125

PR #127, opened 2014-10-02
Add IsoTrack branch and relax the jet multiplicity cutoff for SingleLept...

safarzad:
Add IsoTrack branch and relax the jet multiplicity cutoff in CMGTools/TTHAnalysis/python/analyzers/treeProducerSusySingleLepton.py 

The Lepton RelIso are set to the loose RelISo values to avoid over-cleaning the jet collection.


gpetruc:
No objections from my side, but just to be sure, are you in coordination with @mstoye ?

mstoye:
Hi Giovanni,
We are in contact. As I am watching the changes I would generally say no need to wait for my ok for requests Claudia and Bathool.
Cheers,
Markus

PR #129, opened 2014-10-14
Qvs gandtk met

mariadalfonso:
General interest
1) add Q vs G likelihood
2) fix multilepton config as introduced in 124
3) created metAnalyzer and added tkmet
4) added some official T2tt sample
5) cleanUp of ttHIsoTrackAnalyzer.py

MT2 specific
A) fix initial parameter  in KT
B) switch off with congfigurable AntiKT hemisphere
C) restored default file compression


CMGTools/TTHAnalysis/python/analyzers/ttHJetAnalyzer.py
```diff
@@ -89,8 +88,33 @@ def process(self, iEvent, event):
...
                 if self.testJetID (jet ):
+                    self.computeQGvars(jet)
+                    jet.qgl = self.qglcalc.computeQGLikelihood(jet, rho)
+
+                    #manually match to genparticle
+                    deltaRmin = 999.
+                    foundPartonId = 0
+                    foundPartonMomId = 0
+                    for ipart in event.genParticles:
+                      if ipart.status() != 23: continue
```
gpetruc: I'd say instead _if ipart.status() not in [3,23]:_ so that it works also for pythia 6 (some MC samples are hadronized with that and not pythia 8)

gpetruc:
@pandolf:  Matching with the parton should go in the ttHJetMCMatchAnalyzer.py rather than the ttHJetAnalyzer.py. Can you move it?

Thanks,

Giovanni

pandolf:
on it.

f

mariadalfonso:
the branch has been updated:
moved the parton ID matching in ttHJetMCMatchAnalyzer.py and fixed status and cone

CMGTools/TTHAnalysis/python/analyzers/ttHJetMCMatchAnalyzer.py
```diff
@@ -38,16 +38,29 @@ def isFlavour(x,f):
             if id >  99: return  (id/100)%10 == f
             return id % 100 == f
 
+
+
         event.bqObjects = [ p for p in event.genParticles if (p.status() == 2 and isFlavour(p,5)) ]
         event.cqObjects = [ p for p in event.genParticles if (p.status() == 2 and isFlavour(p,4)) ]
+
+        isParton = isFlavour(p,1) or isFlavour(p,2) or isFlavour(p,3) or isFlavour(p,4) or isFlavour(p,5) or isFlavour(p,21)
+        event.partons   = [ p for p in event.genParticles if ((p.status() == 23 or p.status() == 3) and abs(p.pdgId())>0 and isParton ) ]
```
gpetruc: what's the meaning of isParton here? as it is now, in line 46 it will be defined as boolean computed for the last value of "p" (which is ill defined, but likely it's the last genParticle in the collection since python doesn't have a good scoping and so it probably takes it from the p inside of the list in line 44 ), and then the request of isParton in line 47 it will either accept everything or reject everything.
did you mean instead to define isParton as a function on p to use in the filtering in line 47, i.e.:

```
 isParton = lambda p : isFlavour(p,1) or isFlavour(p,2) or isFlavour(p,3) or isFlavour(p,4) or isFlavour(p,5) or isFlavour(p,21)
 event.partons = [ p for p in event.genParticles if ((p.status() == 23 or p.status() == 3) and abs(p.pdgId())>0 and isParton ) ]
```

also, are you sure you want isFlavour rather than "abs(p.pdgId()) in [1,2,3,4,5,21]" ? as it is, also all hadrons will pass your selection, and similarly would the susy squarks and gluinos.

and, why abs(pdgId()) > 0 ? do we have pdgId zero particles?

pandolf: > also, are you sure you want isFlavour rather than "abs(p.pdgId()) in [1,2,3,4,5,21]" ?
> 
> ^ the above would work. i’ll change it

thanks
f

mariadalfonso:
cleaned up the ttHJetMCMatchAnalyzer.py to follow up the precious comments

PR #143, opened 2014-11-14
Adding new class to recluster jets

clseitz:
New class added to recluster jets with choice of dR and algorithm. 
Ak1.2 jets are formed based on the clean jet collection and added to the standard single lepton ntuples.


gpetruc:
Hi Claudia,

Thanks for contributing the code. However, before merging it I'd ask you if you can improve it slightly.

When working with 4-vectors in cartesian coordinates it's better to use doubles instead of floats to avoid roundoffs when converting from polar to cartesian and vice-versa (it might not matter for fat jets, that have a large mass, but since the code you wrote is general enough to be applied for any fastjet clustering it's better to fix it)

Also, is there a particular reason for using flat vectors of numbers instead of **reco::Particle::LorentzVector** objects in the inteface? 
after all, the fast-jet library doesn't work with plain vectors of coordinates, so you're doing a conversion anyway, if you'd use LorentzVector as intermediate object you could just use the **p4()** method to get it when creating the inpts, and you wouldn't need to convert the output. From the python side, it seems to be possible to create a vector of lorentz vectors via

```
v = ROOT.std.vector(ROOT.reco.Particle.LorentzVector)()
```

i.e. all the necessary dictionaries seem to be in place.

In the next days you can update the pull request using lorentz vectors, or at least change the floats into doubles?

Thanks

Giovanni

gpetruc:
see #149

PR #149, opened 2014-11-24
Reclustering jets (cleaned up version)

gpetruc:
As #143 but with some clean-up of the interface (after an email exchange with @clseitz):
- inputs and outputs are 4-vectors
- no `using namespace` in headers
- output of the analyzer is `event.reclusteredFatJets` instead of just `event.fatJets` (which might otherwise suggest the fat jets from the miniAOD)

Also, added feature to allow making exclusive jets specifying the number or the distance cut via the `getGroupingExclusive` methods.

Validated against #143 on 2000 events of _T1tttt_2J_mGl1500_mLSP100_ (about 4k fat jets), results are identical except for the mass which sometimes changes by 0.01 GeV, which is possibly because of roundoffs due to float precision in the original code.


gpetruc:
@mariadalfonso  can you check if the interface in this version can also serve your usecase?

mariadalfonso:
@gpetruc 
The new interface works also for the hemisphere class (http://dalfonso.web.cern.ch/dalfonso/validation_RND7/mt2_KT.png).

the file that need to be pull requested is this one
https://github.com/mariadalfonso/cmg-cmssw/blob/c6a44925c78543afe1ac2c867cc277061e0e6e84/CMGTools/TTHAnalysis/python/analyzers/ttHTopoVarAnalyzer.py

PR #150, opened 2014-11-25
update to the new KT interface

mariadalfonso:


mariadalfonso:
@gpetruc 
updated the TLorenz vector

PR #161, opened 2014-12-04
implemented photon MC matching as in SMP-13-001

pandolf:


gpetruc:
@mariadalfonso @pandolf : what's the state here? is it everything in and only needs testing from our side?

pandolf:
everything should be in, maria told me she wants to test something right after lunch

f

PR #162, opened 2014-12-04
Update H->tau tau miniAOD setup

steggema:
Note that this 
- modifies a few classes in RootTools, e.g. adds 13 TeV cross sections for the yellow-report parser, and allows for other PU jet ID user floats in the jet-related classes
- touches a few classes in Common and Utilities that are most likely only used by tau-tau to be able to run MVA MET based on miniAOD (otherwise please let me know)
- makes it possible to run PU jet ID on miniAOD by modifying the according classes in RecoJets/JetProducers/src/PileupJetIdAlgo.cc, which should probably be reverted when ported to or merged with newer releases


gpetruc:
Things look ok from susy side. If you're confident from tau tau side, please go ahead merging.

PR #166, opened 2014-12-09
ACDV samples and Component creator

schoef:
ACDV samples added
Component creator: Added fileprefix for local usage.


mariadalfonso:
@schoef Hi Robert, can you do the modification on top of the central release ?

gpetruc:
Hi @schoef,

Ideally, the filePrefix could be derived in a location-independent way with `edmFileUtil -d`. 
I wouldn't advocate doing it once per file, but for example one could do it only once per class of prefix (every `/store/<x>` prefix ).
Could you try to have a look?

PR #168, opened 2014-12-10
IsoTrack and Mt2small fixes

mariadalfonso:
Things of general interest:
1) moved Isotrack to the core (was now cloned in 4 analyzer, default is off)
2) added optional list with overlap check between the isoTrack and IDleptons
3) added options for isotrack treatment for the jet cleaning and in the definition global variables
4) backport of the Photon H/E from the 72X

MT2 related:
1) fixed gamma control variables
2) added variable to the validation script


mariadalfonso:
@gpetruc The IsoTrack and MET are independent now. MetAnalyzer in back after the JetAnalyzer

PR #176, opened 2014-12-12
ACDV samples + fileprefix

schoef:
@gpetruc 
I'll look into your suggestion - could we merge this in the meantime?


mariadalfonso:
@schoef Can you add the cross section to the extra samples (see the other samples in this file) ?

schoef:
Done.

PR #183, opened 2014-12-18
Adapt tau-tau to heppy and include latest MVA MET recipe

steggema:
MVA MET shouldn't affect anyone using the CMG default sparse checkout and supersedes the changes I made to make PU jet ID run in 7.0.6.


mariadalfonso:
Hi Jan.
What you have in CMGTools/H2TauTau/python/proto/physicsobjects/DiObject.py is a duplicate of PhysicsTools/Heppy/python/physicsutils/DiObject.py .

Can you clarify how you run the MVAmet ? (great that you added the tag from the MET group)
In case you add to the events, will be good will be added inside the  Heppy/python/analyzers/objects/METAnalyzer.py
so that is available to everybody ?

Maria

steggema:
Hi Maria,

The DiObject is not an exact duplicate, but relies on the pat::CompositeCandidate instead of the CMG DiObject class as the underlying C++ class. This could in principle be ported to Heppy since the CMG DiObject doesn’t exist anymore, but there may be another user so we’ll have to see.

Concerning MVA MET, we’ve so far been running it as an extra framework step before the final Python analysis. The new tag from the MET group allows us to do this on miniAOD so that we neither have to rely on some precomputed MVA MET nor run on AOD. We’ll have to see whether it makes sense to add something related to the METAnalyzer, so let’s maybe discuss offline.

Jan

PR #193, opened 2015-01-11
update CMGDB pw

gitytakahas:


gpetruc:
Hi Yuta,

Can you make pull requests with this change also for 53X and 70X?

Thanks

Giovanni

PR #207, opened 2015-01-22
Fix for photon 5x5 cluster shapes remaking miniAOD from miniAOD

emanueledimarco:


emanueledimarco:
This is to calculate photon cluster shapes from miniAOD  -> miniAOD in 723_patch1

gpetruc:
Ciao,

I'd suggest a few changes:
1. in the cfg, change the event content to drop the old slimmed photons and keep the new ones (so that by default one can't get those by mistake)
2. change the module name from `slimmedPhotonsFromMiniAOD` to `slimmedPhotons` (so that one does not need to change input tags); of course, the name of the cfi file should remain as it is.

Could you do them and push again to the same branch?

Thanks,

pandolf:
one remark: by talking to emanuele it seems that the methods

def hOVERe(self):
   return self.physObj.full5x5_hadTowOverEm()

def r9(self):
   return self.physObj.full5x5_r9()

would still be bugged. so we’d either have to change 

https://github.com/cms-sw/cmssw/blob/CMSSW_7_4_X/PhysicsTools/Heppy/python/physicsobjects/Photon.py https://github.com/cms-sw/cmssw/blob/CMSSW_7_4_X/PhysicsTools/Heppy/python/physicsobjects/Photon.py

in such a way that the variables are recomputed from the non-bugged showershape variables. or maybe it’s possible to fix also these Photon methods in this fix?

f

gpetruc:
Can you explain why you think they should be bugged?

cbernet:
Hi guys, 

I think that this pull request should instead be sent to the PAT guys, 
as it is not specific to CMG tools, and will be needed by many other people in CMS.
Then CMG can take an updated PAT recipe to get this code. 

Cheers, 

Colin

pandolf:
emanuele told me that he has fixed the reco::Photon::ShowerShape variables, but not the reco::Photon methods. but i’ll let him elaborate

f

emanueledimarco:
@pandolf: exactly. Let me know if I also can change the Photon accessors to fix the other variables which are not directly from the ShowerShape block. 

@gpetruc : So, I can make the changes that Giovanni suggests easily and update the pull request.

gpetruc:
@emanueledimarco @pandolf : looking at the code of r9, it seems to me that once the shower shape block info is updated in the photon the method should return the correct value:
https://github.com/CERN-PH-CMG/cmg-cmssw/blob/CMGTools-from-CMSSW_7_2_3/DataFormats/EgammaCandidates/interface/Photon.h#L207

so, there should be nothing to fix

pandolf:
true: also H/E should work fine.

great then!

f

emanueledimarco:
Done the two changes suggested by @gpetruc

PR #220, opened 2015-01-29
Mt2 tau photon config

mariadalfonso:


CMGTools/TTHAnalysis/python/analyzers/ntupleTypes.py
```diff
@@ -60,6 +60,7 @@
 ##------------------------------------------  
 
 photonTypeSusy = NTupleObjectType("gammaSusy", baseObjectTypes = [ photonType ], variables = [
+    NTupleVariable("genIso",  lambda x : x.genIso, float, mcOnly=True, help="sum pt of all status 1 particles within DeltaR = 0.4 of the photon"),
```
gpetruc: Can you replace `x.genIso` with `getattr(x, 'genIso', -1.0)` ?

Photon MC matching can be turned off at configuration level, since it's quite slow, but in those cases `genIso` is not used.

PR #224, opened 2015-02-03
Heppy - back to using full5x5 clustershapes

pandolf:
now that the photon fix by emanuele is in place, we can go back to the usual 5x5 clustershapes


mariadalfonso:
The Photon.py and PhotonAnalyzer can be improved.
@pandolf Since you are at it and the main customer can you have a look ?

For example the way it's done now, we are overwriting the methods of PAT photons r9 with the r9_5x5.
Also can you move the ID from the PhotonAnalyzer.py to the Photon.py

mariadalfonso:
@pandolf This one should be committed into the branch CERN-PH-CMG:heppy_7_2_2_patch2 and the commit should start with string "Heppy". So like it is, it's hard to be merged.

pandolf:
i’m glad to do whatever is neeeded, i was not aware of these constraints, also because there is no mention of them on the twiki:

https://twiki.cern.ch/twiki/bin/viewauth/CMS/CMGToolsReleasesExperimental#Git_MiniAOD_release_for_PHYS14_C

so lets get this offline and you tell me what to do

f

PR #225, opened 2015-02-03
Include latest SVFit code

steggema:


schoef:
Hi,

after merging this I get an error from here:
https://github.com/CERN-PH-CMG/cmg-cmssw/blob/CMGTools-from-CMSSW_7_2_3/TauAnalysis/SVfitStandalone/interface/SVfitStandaloneMarkovChainIntegrator.h#L22

/afs/hephy.at/scratch/s/schoefbeck/CMS/CMSSW_7_2_3/src/TauAnalysis/SVfitStandalone/interface/SVfitStandaloneMarkovChainIntegrator.h:22:26: fatal error: Math/Functor.h: No such file or directory

Is that something obvious?

Thanks

steggema:
It’s not obvious to me why the package doesn’t compile. Maybe you only get a partial checkout from the merge since the directory is not in your sparse checkout file, so maybe a “git checkout” does the trick (unless you merged with cms-merge-topic, in which case you’d have to remove it by hand, or you want to run the tau-tau analysis).

aelwood:
Hi all,

I just got the same problem and fixed it by adding to .git/info/sparse-checkout the line:

/TauAnalysis/

With the merge it seems like some of the TauAnalysis packages checked out, but not all of them.

Hopefully this information is of use.

Cheers,
Adam

On 9 Feb 2015, at 09:42, Jan Steggemann <notifications@github.com<mailto:notifications@github.com>> wrote:

It’s not obvious to me why the package doesn’t compile. Maybe you only get a partial checkout from the merge since the directory is not in your sparse checkout file, so maybe a “git checkout” does the trick (unless you merged with cms-merge-topic, in which case you’d have to remove it by hand, or you want to run the tau-tau analysis).

PR #232, opened 2015-02-05
Heppy - Replaced PAT photon isolation with RECO isolation

mmasciov:


mariadalfonso:
@mmasciov 
The commit has to happen in the cmg-central/heppy_7_2_2_patch2 branch
The commit comments has to start with the string "heppy"

gpetruc:
Since in any case the target branch of a PR can't be changed once it is made, this can be closed as you'll need do make a new one if you want to pull it into the heppy branch.

PR #233, opened 2015-02-06
Remove CMGTools/Common

steggema:
Removes CMGTools/Common

In addition, remove or adapt the few remaining files that use CMGTools/Common

@gpetruc  @cbotta : Also removes two seemingly outdated files in TTHAnalysis that use Common (running the PAT CMG sequence)


gpetruc:
ok for me, so you can merge when you want

PR #234, opened 2015-02-06
Heppy - Replaced PAT photon isolation with RECO isolation

mmasciov:


mariadalfonso:
@aelwood This PR change the content to the photon isolation (isolation taken from RECO and not PAT now). Please double check.

aelwood:
@mariadalfonso the code looks fine to me. As far as I can see it just adds an extra recoIsolation attribute rather than changing it?

pandolf:
@aelwood be careful: the branches are overwritten here:

https://github.com/CERN-PH-CMG/cmg-cmssw/pull/234/files#diff-0

aelwood:
@pandolf Ah right ok I see. All looks good to me then.

PR #238, opened 2015-02-06
Upstream Heppy updates: logger verbosity and minor fixes

gpetruc:
The only changes not yet in our branch are for the verbosity of the loggers, plus some cleanup, so it should be transparent to everybody, except that if you want the full verbose output you need to run heppy with the "-v" option (or set the default as verbose in your cfg as in c45569e2a9b8afc676fb81b3a5be350c402cd921)

As an added feature, you can pass options from the heppy command line to your cfg. Options can be passed with the syntax `-o flag` or `-o key=value` and can be accessed from the cfg file as in this example fragment

```
  from PhysicsTools.HeppyCore.framework.heppy import getHeppyOption
  test = getHeppyOption('test')
  if test: print "Will run test scenario %r" % test
  if test == '1':
       ....
```


gpetruc:
waiting to hear from @cbernet whether the default should be verbose or quiet, and fix it.

PR #241, opened 2015-02-09
Heppy: Update tau object/analyzer/filler to latest POG recommendations

steggema:


mariadalfonso:
@steggema 
I was cross checking my susy workflow with your PR with ttbar Phy14 sample, I got this error.
What else need to be added ? 

/afs/cern.ch/work/d/dalfonso/CMSSW_7_2_3_feb6/python/PhysicsTools/Heppy/analyzers/objects/TauAnalyzer.pyc in makeTaus(self, event)
     49             tau.idDecayMode = tau.tauID("decayModeFinding")
     50             tau.idDecayModeNewDMs = tau.tauID("decayModeFindingNewDMs")
---> 51             if self.cfg_ana.decayMode and not tau.tauID(self.cfg_ana.decayMode):
     52                 continue
     53 

AttributeError: 'Analyzer' object has no attribute 'decayMode'

steggema:
Thanks, should work now (using correct name + check if the decay mode ID parameter is set in the cfg). I also updated the anti-muon ID discriminator to the latest recommendation.

mariadalfonso:
@steggema 
Things are not backward compatible 
https://dalfonso.web.cern.ch/dalfonso/validation_ttbar_tau/?match=tau_

I think has to do with the tauAntiMuonID and tauAntiElectronID that before weren't in

Maria

steggema:
There are two changes that are backwards incompatible in the default configuration:
1) The dxy and dz calculations now return sensible values; since there are cuts on dxy/dz in the default config, this will have an effect on the selection.
2) Requirement of decayModeFindingNewDMs, which is the minimal ID required by the tau POG and a superset of decayModeFinding. However, this is already required at the miniAOD level, so it should have zero effect on miniAOD.

In addition, if you explicitly set vetoLeptonsPOG to True in your test, the default lepton veto criteria were updated to the latest POG defaults, which I think is reasonable.

If there's a strict requirement for backwards compatibility, we can set the max dxy/dz criteria in the default config to high numbers so that we get exact agreement in the default selection.

mariadalfonso:
Those plots are made with the
https://dalfonso.web.cern.ch/dalfonso/validation_ttbar_tau/?match=tau_

for the 
1) I have those so no selection effect here

tauAna.dxyMax = 99999.
tauAna.dzMax = 99999.

2) Regarding the ID I have the following
    decayModeID = "", # ignored if not set or ""  
    tauID = "byLooseCombinedIsolationDeltaBetaCorr3Hits",
    vetoLeptonsPOG = True,
    tauLooseID = "decayModeFinding",
    tauAntiMuonID = "againstMuonLoose3",  
    tauAntiElectronID = "againstElectronLooseMVA5",   

the  vetoLeptonsPOG introduce the problem likely since we have the tauAntiMuonID and tauAntiElectronID, can we add another flag to apply or not ?

steggema:
If you want to go back to the previous behaviour with the new configuration, you would need to add the non-recommended old anti-electron and anti-muon discriminators (tauAntiMuonID = "againstMuonTight" and tauAntiElectronID = "againstElectronLoose"). I made these configurable with this PR and added the POG recommendations as default.

gpetruc:
Ok,

I think that the way to proceed is:
- put the POG-recommended choices in the default configuration of Heppy
  instead of what's currently there
- for the CMG susy defaults, which are independent from the Heppy ones, I
  would also suggest that we put the POG-recommended values and have people
  override them to the old non-recommended value in their cfg.
  
  Giovanni

gpetruc:
@mariadalfonso can you check that after setting also those last two parameters to the non-recommended old value then everything checks clean? (and when done, mark as validated)

pandolf:
@steggema: as soon as you've done it, tell us what we need to put in our cfgs to get back to the latest selection. (btw: thanks!)

steggema:
@gpetruc: The PR as of yesterday should contain the latest POG defaults, so it's good to go from my side. Note that the electron/muon vetoes are disabled by default since you may generally not want to apply them (in particular the electron veto) if the analysis vetoes hadronic taus. We can discuss this offline if people are interested in more details.

@pandolf: If you want to get to the previous (?) selection, you'd need to set dzMax to a large value (though in principal you'd only want a tau that comes from the PV used in your analysis), and you'd need to set the anti-mu/anti-e discriminator names to the ones mentioned above. About a reasonable selection for a specific analysis, we can probably better discuss offline, but the new default selection is fairly loose so that it's likely that it can serve a high number of analyses.

mariadalfonso:
I wiill complete the test and modify the susy core this afternoon

mariadalfonso:
@gpetruc @steggema 
I made a PR to adjust the susyCore parameters and restore the old value for the MT2
https://github.com/CERN-PH-CMG/cmg-cmssw/pull/248

PR #249, opened 2015-02-12
Heppy: jets and photons cleaning and met type1 with chs jets

mariadalfonso:


PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
```diff
@@ -115,7 +115,10 @@ def process(self, event):
         ## Clean Jets from photons
         photons = []
         if hasattr(event, 'selectedPhotons'):
-            photons = [ g for g in event.selectedPhotons ]
+#            photons = [ g for g in event.selectedPhotons ]
+            for g in event.selectedPhotons:
+              photons.append( g )
+              break
```
gpetruc: `photons = event.selectedPhotons[:1]`

And, please make if configurable whether onlythe first photon is used or not.

PR #260, opened 2015-02-17
H2TauTau: added em channel, revised a bunch of analysers, uniformed the ntuple prod across the four channels

rmanzoni:
Added em channel.

Uniformed the four channels: mostly implied a cleanup of the tt sequence and hence TauTauAnalyzer, and in addition we created a base HTT (H2TauTauTreeProducer) tree producer class, from which the channel-specific producers inherit.

Modified the generator level related analysers, so that the job of creating the separate gen particle collections is done only once by Heppy GeneratorAnalyzer (now included in every HTT sequence).


CMGTools/H2TauTau/python/proto/analyzers/GenErsatzAnalyzer.py
```diff
@@ -34,8 +35,8 @@ def process(self, event):
             return True
         result = super(GenErsatzAnalyzer, self).process(event)
         
-        event.genZs = [ZBoson(part) for part in event.genParticles if part.status()==3 and part.pdgId()==23]
-        event.genWs = [WBoson(part) for part in event.genParticles if part.status()==3 and abs(part.pdgId())==24]
+        event.genZs = [ZBoson(part) for part in event.genVBosons if part.pdgId()==23]
+        event.genWs = [WBoson(part) for part in event.genVBosons if part.pdgId()==23]
```
steggema: Shouldn't this be abs(part.pdgId()) == 24?

CMGTools/H2TauTau/python/proto/analyzers/H2TauTauTreeProducer.py
```diff
@@ -0,0 +1,434 @@
...
+    def fillDiLepton( self, tree, diLepton):
+        self.fill(tree, 'visMass'  , diLepton.mass()     )
+        self.fill(tree, 'svfitMass', diLepton.svfitMass())
+        self.fill(tree, 'pZetaMET' , diLepton.pZetaMET() )
+        self.fill(tree, 'pZetaVis' , diLepton.pZetaVis() )
+        self.fill(tree, 'pZetaDisc', diLepton.pZetaDisc())
+        self.fill(tree, 'mt'       , diLepton.mTLeg2()   )
+        self.fill(tree, 'mtleg2'   , diLepton.mTLeg2()   ) # RIC: handy for tt, redundant but non destructive
+        self.fill(tree, 'mtleg1'   , diLepton.mTLeg1()   )
+        # RIC: can't find cov matrix... will look it up later
```
steggema, line 217: should be in the MET object

CMGTools/H2TauTau/python/proto/analyzers/GenErsatzAnalyzer.py
rmanzoni: yes it should, it must be a typo of mine

CMGTools/H2TauTau/python/proto/analyzers/H2TauTauTreeProducer.py
rmanzoni, line 217: diLepton.met().getSignificanceMatrix() crashes.
Agreed to go ahead without this is and take not that this needs to be fixed.

PR #263, opened 2015-02-18
Heppy - recover original MT2 precision

brucala:
title says it all

### DIFF
--- a/PhysicsTools/Heppy/src/Davismt2.cc
+++ b/PhysicsTools/Heppy/src/Davismt2.cc
@@ -42,7 +42,7 @@ namespace heppy {
 /*The user can change the desired precision below, the larger one of the following two definitions is used. Relative precision less than 0.00001 is not guaranteed to be 
   achievable--use with caution*/ 
 
-const float Davismt2::RELATIVE_PRECISION = 0.0001;
+const float Davismt2::RELATIVE_PRECISION = 0.00001;
 const float Davismt2::ABSOLUTE_PRECISION = 0.0;
 const float Davismt2::ZERO_MASS = 0.0; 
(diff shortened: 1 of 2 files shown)


gpetruc:
does it make any significant difference in the timing?
@mariadalfonso

pandolf:
hi guys

just to inform you that while it would be indeed interesting to see if this affects timing, we don't really have a choice: this is a bug that was introduced in november in heppy [1], of which we were not aware, and we need to go back to the previous value for syncronization reasons.

ciao
f

[1]: as you can see here:

https://github.com/CERN-PH-CMG/cmg-cmssw/commi/908095816aa7214b8e9b143cb7f365600b2e0d59

when moving the parameter from Davismt2.h and mt2w_bisect.h to Davismt2.cc and mt2w_bisect.cc a zero was "lost"

pandolf:
sorry, missing t in the link:

https://github.com/CERN-PH-CMG/cmg-cmssw/commit/908095816aa7214b8e9b143cb7f365600b2e0d59

gpetruc:
I see, maybe next time write this more clearly in the description of the PR?

Also, why don't you fix also the other file then?

mariadalfonso:
@gpetruc 
the module ttHTopoVarAnalyzer that call this class it account only for 1% of the overall time in the run_susyMT2_cfg.py sequence 
So with Bruno fix, the picture doesn't change

gpetruc:
Ok, so I'll just wait for the  fix also in `mt2w_bisect` and then merge the branch

pandolf:
@brucala can you do this?

brucala:
done

PR #269, opened 2015-02-19
Heppy: Fix issues in gen analyzer and leptons

steggema:
Gen analyzer: Account for weird decay chains in Phys14 ggH samples (where status 3 Higgs decays to identical status-2 copy of itself + its other decay produts)
Leptons: Remove outdated H->tau tau specific electron and methods in lepton; remove or rename a few duplicated function definitions

Note: Should be transparent unless you used tau-tau-specific isolation methods (all charged iso) (apart from the gen analyzer fix)


mariadalfonso:
@gpetruc This PR it's transparent for the MT2 configuration

PR #271, opened 2015-02-19
H2TauTau: updates 

rmanzoni:
_ prepared the htt analyzer to modified Heppy objects as in Jan's PR https://github.com/CERN-PH-CMG/cmg-cmssw/pull/269
_ made sure that Trigger analyzer works for all channels
_ fixed flaws in htt analyzers
_ added an option to produce all channels in one go in one file
_ added back PFcandidates in event content


steggema:
The history we would import is a bit excessive for the number of actually changed files. Maybe it makes more sense to re-apply the changes on top of a clean checkout?

rmanzoni:
Yes, makes sense.
In this case I may want to wait for your PR to be merged and avoid yet another intermediate step.

PR #277, opened 2015-02-24
CMG Tools branch for MonoX analyses (first trivial shot)

emanueledimarco:
This PR also includes a PR to heppy branch previously done (by cherry-picking files).
@avartak @bachtis also to follow this PR


gpetruc:
Hi Emanuele,

You have deleted files from the release (at least in commit 4f4b45d), which is **not** what you wanted to, you should just have removed them from the sparse checkout (if they had ended there for any reason) and do `git read-tree -mu HEAD` (or anything similar) to get rid of them from your working copy **without** deleting them from the repository.

I will close this pull request now.

What you should do is create a new branch and copy there the good commits (with cherry-pick or `rebase --interactive`), and make a new pull request out of it. If there's a commit that contains both the deletes and something good, you can get its differences e.g. with `cherry-pick -n` and then use `git reset` appropriately do commit only the right things)

PR #279, opened 2015-02-25
CMG Tools branch for MonoX analyses

emanueledimarco:
Second attempt after putting back Configuration in the release...
@avartak @bachtis @lsoffi also to follow this PR


CMGTools/TTHAnalysis/python/analyzers/ttHAlphaTVarAnalyzer.py
```diff
@@ -1,3 +1,6 @@
+from PhysicsTools.Heppy.analyzers.core.Analyzer import Analyzer
+from PhysicsTools.HeppyCore.utils.deltar import deltaR, deltaPhi
```
aelwood, line 2: @emanueledimarco These lines are the same as 11 and 12? This addition isn't necessary I don't think

emanueledimarco, line 2: yes you are right. I removed them (sorry I had changed those files before you did the migration to heppy and I did not notice the repetition)

PR #280, opened 2015-02-25
add taus in susyMultilpeton analysis + electron eta SC + extra jet variables for SUSY

mmarionncern:
This PR adds :
- the full tau analyzer in the susyMultilepton analysis
- the counter histogram in the output file
  -the electron eta supercluster in the additional variables for leptons
- plug per default the jets extra variables for SUSY analyses (useful for jetId and potential handling of noisy jets in data)

(In case it is not the proper procedure for a PR, let me know)

### DIFF
--- a/CMGTools/TTHAnalysis/cfg/run_susyMultilepton_cfg.py
+++ b/CMGTools/TTHAnalysis/cfg/run_susyMultilepton_cfg.py
@@ -89,6 +89,10 @@ treeProducer = cfg.Analyzer(
      collections = susyMultilepton_collections,
 )
 
+## and the tau producer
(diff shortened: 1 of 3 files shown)


gpetruc:
Hi Matthiew,

You should not mix changes to Heppy files with changes to CMGTools file in the same commit.
Can you make another PR with those changes split in separate commits?

In addition, we would also  a second pull request with just the Heppy change, made on the heppy branch (which you can make by cherry-picking that commit)

```
 Giovanni
```

mmarionncern:
Ok, sorry for that, will prepare separated PRs then.

PR #281, opened 2015-02-25
add taus in susyMultipleton analysis  + extra jet variables for SUSY

mmarionncern:
This PR adds :
- the full tau analyzer in the susyMultilepton analysis
  -  the counter histogram in the output file in the same analysis
  -   plug per default the jets extra variables for SUSY analyses (useful for jetId and potential handling of noisy jets in data)
- add 30 GeV b-jet counters

### DIFF
--- a/CMGTools/TTHAnalysis/cfg/run_susyMultilepton_cfg.py
+++ b/CMGTools/TTHAnalysis/cfg/run_susyMultilepton_cfg.py
@@ -89,6 +89,10 @@ treeProducer = cfg.Analyzer(
      collections = susyMultilepton_collections,
 )
 
+## and the tau producer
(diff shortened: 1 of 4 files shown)


gpetruc:
You have again mixed changes to heppy and cmgtools in the same commit.

And, i think it would be better to take the tau analyzer from the cmg susy config, with replaces if needed, as that would go in the direction of eventually having consistent taus across the different analyses.

And, we can discuss it but i don't think i'd rather change the default output mode of multilepton trees.

mmarionncern:
Humm wrong branch, sorry for that. Will fix it tomorrow.

The change of output mode allows to get an histogram with a generated event counter in it. This could be discussed, but the functionality is for my point of view really interesting to not be used.

PR #289, opened 2015-03-02
 Wmass 5.3.22 february

mariadalfonso:
This PR contains:
1) update to the code to derive recoil corrections in since with AN-15-013 v3
2) update to the wmass fits (handling of muon, recoil corrections)


gpetruc:
Given it affects only the wmass, you can merge when you want.

I took away the "local" flag, as that's supposedly meant for distinguishing something that is local to a subset of the area it is flagged to (e.g. something that affects only MT2 or AlphaT and not the whole "susy/tth" flag), here it seems to me it's modifying all Wmass.

PR #294, opened 2015-03-10
MTtwo additions

mariadalfonso:
This contains changes and additions for the MT2


gpetruc:
@aelwood: in AlphaT you don't use the  ttHTopoVarAnalyzer.py, right? (or if you do, you're ok with the change?)

Except for this question, @mariadalfonso you can merge the PR when you want (I'm assuming you tested it on the MT2 side; for multilepton side I checked that it is transparent)

gpetruc:
I'll do the merge since it now conflicts with #296 due to the moved validation plot, and I have to commit anyway some fix to the validation script from our side.

aelwood:
@gpetruc we don't use the ttHTopoVarAnalyzer.py for any variables required by the analysis so any changes here are fine for us.

PR #295, opened 2015-03-10
little changes to DEFAULT SUSY LEP ANALYZER configuration

cbotta:
these pull request contains two little changes to the default lepAnalyzer configuration in sysyCore. the electron pseudorapidity acceptance cut is now 2.5 (no reasons to cut at 2.4) and the muon-electron cross cleaning (with muon priority) uses a dR cone of 0.05. this cut value is the one chosen by the HZZ4l group. dR =0.02 was not supported by any study.

### DIFF
--- a/CMGTools/TTHAnalysis/python/analyzers/susyCore_modules_cff.py
+++ b/CMGTools/TTHAnalysis/python/analyzers/susyCore_modules_cff.py
@@ -160,7 +160,7 @@ lepAna = cfg.Analyzer(
     # loose electron selection
     loose_electron_id     = "POG_Cuts_ID_2012_Veto_full5x5",
     loose_electron_pt     = 7,
-    loose_electron_eta    = 2.4,
+    loose_electron_eta    = 2.5,
     loose_electron_dxy    = 0.05,
     loose_electron_dz     = 0.1,
     loose_electron_relIso = 0.5,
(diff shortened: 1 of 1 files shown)


pandolf:
even if indeed "little" this Pull Request is not backwards compatible, and should therefore be publicized **_MORE STRONGLY**_ to the entire group. all analyzers should be aware that their results might change!!

as for MT2, it seems to me that we are not affected, as we re-assign the electron eta cut automatically:

https://github.com/CERN-PH-CMG/cmg-cmssw/blob/CMGTools-from-CMSSW_7_2_3/CMGTools/TTHAnalysis/cfg/run_susyMT2_cfg.py#L24

@mariadalfonso, could you check that this is indeed the case? (ie that we are not affected by this PR?)

gpetruc:
Hi @pandolf,

According to the policy we have set, it's enough for the pull request to be flagged as susy/tth (meaning it affects everybody taking from susy core config), and to not be flagged as local, enhancement or transparent. That is, what's being done is sufficient.

However plese feel free to take care of hanging a huge billboard in front of B40 to publicize the change, have it advertized at the WGM, and maybe get some coverage live on BBC news and on the front page of the Tribune de Geneve.

gpetruc:
`Merging in the core. Differences are quite tiny in truth (0.5%, http://gpetrucc.web.cern.ch/gpetrucc/drop/plots/ttH/72X/validation_pr295/ ) 
@mariadalfonso @pandolf @aelwood @safarzad you can see whether you want to change something in your analyzer in case you were taking the configuration from the susy-core and don't like the change.

PR #296, opened 2015-03-10
clean up of plotter dir: susy-multilepton, susy-mT2, ttH-multilepton

cbotta:
this pull request contains a cleanup of the /plotter/ directory. the scripts to use the tools to make plots, compute yileds, make datacards ecc... that belong to each analysis are stored in the following subdirectories: tth-multileptons, susy-multileptons, susy-mT2. Feel free to add additional directories if your analysis-group makes use of these tools.


gpetruc:
flagging as local since it only affects the plotter files currently in git, which for the moment are only used in ttH/susy-multilepton and not yet from other analysis

PR #298, opened 2015-03-11
Adding indexes of jets sorted by btagCSV (iJetByCSV) computed in ttHLepEventAnalyzer.

cirkovic:


gpetruc:
This way doesn't work, you can't add in Heppy something that is computed in CMGTools.
Please add instead this variable in the CMGTools `ntupleTypes` .

Also, is this adding this variable to the jets really useful? shouldn't it be better to add a list of indices, which then can be used to access the jets, a bit like done in https://github.com/CERN-PH-CMG/cmg-cmssw/commit/479895166e153ff6ef32d7827ed0f1d0c90e69df#diff-934490a7730e0bdb9b6cff3c07f52e59

PR #299, opened 2015-03-11
Adding indexes of jets sorted by btagCSV (iJetByCSV) computed in ttHLepEventAnalyzer.

cirkovic:


gpetruc:
see my comment on #298 about indices

PR #306, opened 2015-03-17
Heppy: new ele ID WPs     H2TauTau: updates towards PHYS14 

rmanzoni:
Various updates to H2TauTau in preparation of the full production of Phys14 samples:
- full SVfit information saved in flat trees
- got rid of SVfit warnings by fixing the ele and muon masses to their nominal values
- connect.py now accepts the new htt_fullsel_tree_CMG*root filename (contains mt,et,tt,em at once)
- new way to retrieve information about the components. Created httConnector class, cleaned up samples directory, reduced the number of database calls the minimum (better use of cached info and avoids jobs to be killed when the number of DB calls tops the max). Actually this is not final, as we should save more info (e.g. computing efficiency of the parent dataset) into the pickle files
- a few bug fixes 

Updated Heppy Electron.py to include Phys14 cut based ID working points. @gpetruc @cbotta @steggema
CSA14 cut based ID WPs have been updated too, as I found that weren't on par with the latest prescriptions on the TWiki 
https://twiki.cern.ch/twiki/bin/viewauth/CMS/CutBasedElectronIdentificationRun2#Working_points_for_CSA14_samples?rev=13


steggema:
The H->tau tau part is fine
@gpetruc can be merged if the heppy part is fine

gpetruc:
Looks like nobody else is using the CSA14 electron ids (at least not in the cfgs committed to git), so the heppy part is likely transparent to everybody.
Merging.

PR #313, opened 2015-03-20
Remake pull request 310

aelwood:
Added option to Jet Analyzer that allows gen jets to be cleaned from reco photons.

Added matched mcPt information into the tree for leptons and photons in a way analogous to jets.

Fixed problem with the alphaT variable analyzer after the RootTools cleanup

Make HT, MHT and alphaT from gen jets as well.


aelwood:
I'll check the slow down from the extra Gen variables in a bit. I doubt it will be significant however

gpetruc:
Hi,

Can you also make a second PR on the heppy branch with just the two heppy commits (I beileve they're 4c7e4c1m 49357da )

Thanks

aelwood:
@gpetruc done https://github.com/CERN-PH-CMG/cmg-cmssw/pull/315

PR #317, opened 2015-03-20
Heppy PdfWeight: moved from CMGTools

mariadalfonso:


gpetruc:
I have checked that it compiles, but I'd need a valid configuration of it to see whether it runs.

PR #318, opened 2015-03-22
Boris cherry picked for cmg

mangano:
In addition to changes affecting only MT2 cfg file, the following has been changed:
- Changed names of fixPhoton components to match the names of the original components
- Changed QCD_Pt120to170 component to use the same type of input miniAOD dataset as the other QCD_Pt components (this guy was probably not ready when sample py was originally modified)
- Added PHYS14_V4_MC JEC txt files


mariadalfonso:
@aelwood please note the changes of the fixphoton datasets
@gpetruc you can merge

PR #322, opened 2015-03-25
FatJet from AK4

safarzad:
Updating FatJet from AK4 for Heppy, and including it to the singlelepton cfg/producer

### DIFF
--- a/CMGTools/TTHAnalysis/python/analyzers/ttHReclusterJetsAnalyzer.py
+++ b/CMGTools/TTHAnalysis/python/analyzers/ttHReclusterJetsAnalyzer.py
@@ -1,7 +1,11 @@
-from CMGTools.RootTools.fwlite.Analyzer import Analyzer
-from CMGTools.RootTools.fwlite.AutoHandle import AutoHandle
+from PhysicsTools.Heppy.analyzers.core.Analyzer import Analyzer
+from PhysicsTools.HeppyCore.framework.event import Event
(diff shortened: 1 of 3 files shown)


CMGTools/TTHAnalysis/python/analyzers/ttHReclusterJetsAnalyzer.py
```diff
@@ -12,14 +16,14 @@ def declareHandles(self):
...
-        super(ttHReclusterJetsAnalyzer,self).beginLoop()
+    def beginLoop(self, setup):
+        super(ttHReclusterJetsAnalyzer,self).beginLoop(setup)
         self.counters.addCounter('pairs')
         count = self.counters.counter('pairs')
         count.register('all events')
 
     def makeFatJets(self, event):
-        objects40jc = [ j for j in event.cleanJets if j.pt() > 40.0 and abs(j.eta())<2.4 ]
+        objects40jc = [ j for j in event.cleanJets if j.pt() > 30.0 and abs(j.eta())< 5.0 ]
```
gpetruc: while it's technically correct, maybe it would be cleaner not to call "objects40jc" jets that are 30 GeV and no longer central.
Also, for this case I think the cleanest solution is to start from `event.cleanJetsAll` make the p<sub>T</sub> and |&eta;| cuts configurable in the `ttHReclusterJetsAnalyzer`.
Could you do it and update the PR?

clseitz:
Thanks for pointing this out, we made it out now more configurable.

PR #323, opened 2015-03-25
Heppy: Safe printing for particles that are directly put into the event

steggema:
Fixes crash when particle is directly put into the event but daughters are not available (and it anyway makes more sense to not print the daughters)

Will be added to another PR for the CMG branch.

### DIFF
--- a/PhysicsTools/HeppyCore/python/framework/event.py
+++ b/PhysicsTools/HeppyCore/python/framework/event.py
@@ -37,6 +37,7 @@ class Event(object):
             if isinstance( value, collections.Iterable ) and \
                    not isinstance(value, (str,unicode)) and \
                    not isinstance(value, TChain) and \
+                   not hasattr(value, 'numberOfDaughters') and \
                    not recursive :
                 tmp = map(str, value)
 
--- a/PhysicsTools/HeppyCore/python/framework/event.py
+++ b/PhysicsTools/HeppyCore/python/framework/event.py
@@ -1,4 +1,3 @@
-import collections
 from ROOT import TChain
 
 class Event(object):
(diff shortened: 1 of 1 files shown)


gpetruc:
Uhm, this way seems more of a hack to me. I think it should be the `str` method of the object to be fixed to not print daughters if not available.
The event here should not need to know about the `numberOfDaughters` method...

gpetruc:
Maybe you can discuss it also with @cbernet @arizzi

cbernet:
Hi guys, 

I agree with Giovanni. 
HeppyCore is only for the bare framework. 

Cheers, 

Colin

steggema:
I think the meaning of the printout of everything that is iterable is to print out all components of objects that are container-like. I now see the point that heppy core shouldn't know about CMS-internals, but I can't come up with a nice solution. Maybe someone else can.

The issue is that all CMS particles are iterable (even if they're not containers, which is maybe the core of the issue), but the iteration may fail, which means that it's not possible to e.g. put a certain generator particle into the event in default heppy running. As you know, the event is printed a number of times by default, so this leads to a seg fault that's not very obvious.

Specifically @gpetruc : The **str** method is ok, heppy explicitly iterates on the particle in addition to this and tries to print everything that's "in" the particle.

arizzi:
btw what is the point of the recursive iteration in HeppyCore? shouldn't we just use repr vs str to handle how to represent contained objects in containers?

steggema:
I think I have a solution that works:

replace 
  isinstance( value, collections.Iterable )
with
  hasattr(value,  '**contains**')

which probably also makes more sense since that would only print the objects in actual containers and not of everything that's iterable.

I'll update the PR.

cbernet:
Hi Jan, 

What exactly do you mean by «containers» ? CMSSW containers? 
We actually want to print the objects inside user-defined python lists attached to the event, for example. 

Cheers, 

Colin

steggema:
Hi Colin,

I mean python containers, which are defined by implementing the **contains** method. We may equally check for "isinstance(value, collections.Container)" if that's clearer. The updated PR will print the contained objects for everything that's such a container but not for everything that's a iterable (as in the case of CMSSW gen particles, that may lead to crashes). To me this makes sense as the printout should print what's in turn contained in the event, and not necessarily what objects return when iterated over.

Cheers,
Jan

cbernet:
Hi Jan, 

Sounds good, thanks for the clarification. 
Cheers, 

Colin

cbernet:
@gpetruc sorry I should have let you merged this. I thought it was my repo.

PR #326, opened 2015-03-26
Heppy: move meduim muon ID from LeptonExtraType to  LeptonType

safarzad:
Heppy: move meduim muon ID from LeptonExtraType to  LeptonType


gpetruc:
Hi,

Can you also make an equivalent pull request into the heppy branch? 
(this one is in the CMGTools branch)

Flagging as enhancement, as it only adds one variable (and nothing at all for people that start from `leptonTypeExtra`)

Thanks,

Giovanni

PR #327, opened 2015-03-27
Changes required to run TTHAnalysis fw over crab; new features for photon objects

mangano:
New features for photon objects from @pandolf:
- implemented randomCone charged isolation for photons
- now saving photon charged isolation both with deltaR 0.3 and 0.4
- now saving photon gen iso both with deltaR 0.3 and 0.4

Cleaning of jetSusyExtra from @pandolf 
- removed useless branches (replicas) from jetSusyExtra 

From @brucala:
- Added remote copy (so far to PSI only) feature into batchmanager
- MagneticField/ParametrizedEngine package taken from cmssw 73X in order to run TTHAnalysis over crab
- Few fixes in CMG to build MagneticField/ParametrizedEngine correctly 
- Few fixes in heppy to correctly use paths containing environmental variable 
- MT2-specific CRAB tools under CMGTools/TTHAnalysis/cfg/crab 

NB: At the same time as accepting this PR, the file 

/afs/cern.ch/user/c/cmgtools/public/sparse-checkout_72X_heppy 

should be modified to include also the following line:

/MagneticField/ParametrizedEngine/


PhysicsTools/Heppy/python/analyzers/objects/PhotonAnalyzer.py
```diff
@@ -182,5 +278,7 @@ def process(self, event):
     etaMax = 2.5,
     gammaID = "PhotonCutBasedIDLoose",
     do_mc_match = True,
+    do_randomCone = False,
```
gpetruc, line 281: @mangano @pandolf : in addition to adding the default here, you have to add it also to `susyCore_modules_cff.py`.
Can you add a commit for this and update the PR?

mangano, line 281: Hi Giovanni, 

Fix committed. 

Boris

PR #328, opened 2015-03-27
Protection for running on data in some analyzers and few monoX additions

emanueledimarco:
The protection is only in the analysers to compute topological variables in TTHAnalysis and in the analogous one in MonoXAnalysis


gpetruc:
@mariadalfonso : I think this counts as bugfix as the change in the topVarAnalyzer should only be the addition of the if to avoid computing gen quantities on data, but please double-check

PR #330, opened 2015-03-27
Heppy: Add gen tau decay mode finding + save gen-matched tau in reco tau

steggema:


gpetruc:
This is the Heppy part of #329, right ?

steggema:
Yes, exactly.

gpetruc:
Ok, from my side you can merge both when you want.

PR #334, opened 2015-03-31
H ZZ 4l updates, mainly in FSR recovery

gpetruc:
The only change outside the H&rarr;ZZ&rarr;4l subsystem beyond PR #342 is the addition of a new method for FSR-aware isolation to electrons and muons `absIsoWithFSR`, more in line with the other Heppy isolation methods (eventually, the older methods `absIso` and `absIsoWithEA` from the `Lepton` class can be also removed, if nobody uses them; I think H&rarr;ZZ&rarr;4l might have been the only customer). 
So, it should be transparent to everybody.


gpetruc:
@bachtis

PR #335, opened 2015-03-31
Adding LHEEventProduct information (weights for scales, pdfs)

jmvizan:
Adding LHEEventProduct information. For the moment it is only added to the Susy Multilepton tree


mariadalfonso:
Hi @jmvizan . I tried your PR put of the box. But it crash for samples w/o LHEEventProduct

Exception: 
            Cannot find collection with:
            type = LHEEventProduct
            label = source

Can you modify the code such that returns an empty list. Something similar was done here.
https://github.com/CERN-PH-CMG/cmg-cmssw/blob/CMGTools-from-CMSSW_7_2_3/CMGTools/TTHAnalysis/python/analyzers/susyParameterScanAnalyzer.py

jmvizan:
Hi Maria.
Thanks a lot for the comments. The problem I see is that with my miniAOD's with the LHE branch (you can see a dump of their event content here):
/afs/cern.ch/work/v/vizan/public/tmp/TTZToLL_aMCatNLO_diLepFilterSS.MINIAODSIM00.root ->miniAOD
/afs/cern.ch/work/v/vizan/public/tmp/log_store.log -> corresponding log of edmEventSize -v
I see that self.mchandles['lhe'].isValid() is set to False and I get the corresponding "ERROR: Missing LHE header in file" when I run the susyParameterScanAnalyzer module.

In addition, in
https://github.com/CERN-PH-CMG/cmg-cmssw/blob/CMGTools-from-CMSSW_7_2_3/PhysicsTools/Heppy/python/analyzers/gen/GeneratorAnalyzer.py#L57

If I add some prints in def makeMCInfo(self, event):

```
def makeMCInfo(self, event):
    verbose = getattr(self.cfg_ana, 'verbose', False)
print "Before rawGenParticles self.mchandles['genParticles'].isValid()=",self.mchandles['genParticles'].isValid()
    rawGenParticles = self.mchandles['genParticles'].product()
print "After rawGenParticles self.mchandles['genParticles'].isValid()=",self.mchandles['genParticles'].isValid()
```

I get as output the first time I reach that point:
Before rawGenParticles self.mchandles['genParticles'].isValid()= False
After rawGenParticles self.mchandles['genParticles'].isValid()= True

But I thought that you require the isValid() condition to avoid to call the .product() method when the info is not present. So I guess I'm missing something about how the logic to use this protection works,
Jesus

gpetruc:
Did you set also the `mayFail = True` in the handle?

jmvizan:
Hi, I double checked, and in fact yes, I have used the mayFail=True (as it is in susyParameterScanAnalyzer, where it also doesn't seem to work the readout of the LHE branch). In fact I see the same behaviour for genParticles and LHEweights (in one case mayFail is to true and in the other not specified). In other words for this snippets:

```
 def declareHandles(self):
     super(GeneratorAnalyzer, self).declareHandles()
     self.mchandles['genParticles'] = AutoHandle( 'prunedGenParticles', 'std::vector<reco::GenParticle>' )
```
- ```
    self.mchandles['LHEweights'] = AutoHandle( 'source', 'LHEEventProduct', mayFail = True )
  ```
- ```
    print "Before rawGenParticles self.mchandles['genParticles'].isValid()=",self.mchandles['genParticles'].isValid()
   rawGenParticles = self.mchandles['genParticles'].product() 
  ```
- ```
    print "After rawGenParticles self.mchandles['genParticles'].isValid()=",self.mchandles['genParticles'].isValid()
  ```
- ```
    print "beforeLHEproduct self.mchandles['LHEweights'].isValid()=",self.mchandles['LHEweights'].isValid()
  ```
-       temp=self.mchandles['LHEweights'].product()
-       print "afterLHEproduct self.mchandles['LHEweights'].isValid()=",self.mchandles['LHEweights'].isValid()

I got the following output the first time I reach the  prints:
Before rawGenParticles self.mchandles['genParticles'].isValid()= False
After rawGenParticles self.mchandles['genParticles'].isValid()= True

beforeLHEproduct self.mchandles['LHEweights'].isValid()= False
afterLHEproduct self.mchandles['LHEweights'].isValid()= True

after that I always get for all the print statements *.isValid= True

I guess this is not the expected behaviour, do you have any guess what might be causing it? Thanks a lot

gpetruc:
Ah ok, 

I see, this could be related with the change from @arizzi in Heppy to make handles load the things only when the `product()` method is called (see https://github.com/CERN-PH-CMG/cmg-cmssw/commit/58658ddf172b2e57e204ac538c5e905f80a2d30f#diff-9108d0025108b6718f2574bb63b7b413 )

Can you test if calling `self.mchandles['LHEweights'].ReallyLoad(event.input)` right after `self.readCollections( event.input )` makes thing work correctly? (i.e. the handle is reported as valid if and only if the object is present, already before calling `product()` )

If yes, than we can consider to add to `AutoHandle` an extra keyword argument `lazy = True` (which we instead set as `False` in cases like this), and if `lazy` is not true then `Load` will already call `ReallyLoad`.

Giovanni

arizzi:
I think the issue is not with the AutoHandle, please see:
https://hypernews.cern.ch/HyperNews/CMS/get/edmFramework/3430.html

gpetruc:
Ciao,

Yes, I think I had been aware of the bug in the past since I see that the
AutoHandle does in recreate the Handle after a failure
https://github.com/CERN-PH-CMG/cmg-cmssw/blob/CMGTools-from-CMSSW_7_2_3/PhysicsTools/Heppy/python/analyzers/core/AutoHandle.py#L35
So, I think the code should work as is in 72X (provided the handle is not
lazy) and in 74X we can benefit from Dan's fix in
https://github.com/cms-sw/cmssw/pull/7502 and avoid re-creating things
after a failure.

```
Giovanni
```

jmvizan:
Thanks for checkin up on this. Let me just mention that adding self.mchandles['LHEweights'].ReallyLoad(event.input) after

def process(self, event):
  self.readCollections( event.input )

as suggested by Giovanni in fact seems to help (isValid is set to True before calling .product()).

gpetruc:
Hi,

If adding the ReallyLoad works, then can you implement the cleaner solution in the previous messages, ie making the lazyness of the AutoHandle configurable (with default true, as now), set this and the susy parameter scan analyzer to be not lazy, and update this pull request (for the latter, you just need to push the changes again on the same branch of your github)

Thanks

jmvizan:
Alright. I was not sure if you wanted to implement this AutoHandle solution yourself. I will take a look to it, and let you know if I have some problems. Thanks

jmvizan:
Hi again, I have just implemented the "lazy" approach suggested by Giovanni. Everything work fine now, at least for my samples, for which the LHE branch is present (I don't have miniAOD without this branch to test, but in principle it should work),
Jesus

PhysicsTools/Heppy/python/analyzers/gen/GeneratorAnalyzer.py
```diff
@@ -53,10 +56,12 @@ def __init__(self, cfg_ana, cfg_comp, looperName ):
         self.makeAllGenParticles   = cfg_ana.makeAllGenParticles
         self.makeSplittedGenLists  = cfg_ana.makeSplittedGenLists
         self.allGenTaus            = cfg_ana.allGenTaus if self.makeSplittedGenLists else False
+	self.makeLHEweights  = cfg_ana.makeLHEweights
  
     def declareHandles(self):
         super(GeneratorAnalyzer, self).declareHandles()
         self.mchandles['genParticles'] = AutoHandle( 'prunedGenParticles', 'std::vector<reco::GenParticle>' )
+	self.mchandles['LHEweights'] = AutoHandle( 'source', 'LHEEventProduct', mayFail = True, lazy = False )
```
gpetruc: maybe you could declare the handle only if `self.makeLHEweights` is true?

gpetruc:
Hi,
If you can fix also the last thing I've commented on, not declaring at all the `self.mchandles['LHEweights']` if `self.makeLHEweights` is false, then I can merge this PR.
I've already tested that it works fine both on LHE events and on pythia ones.
Giovanni

jmvizan:
Excellent! I just applied the fix you mention (define the handle only if the makeLHEweights is True). I tested both with makeLHEweights set to True or False and seems to work fine,
Jesus

PR #336, opened 2015-04-02
Leptons update 

mariadalfonso:
Heppy Lepton Analyzer: 
1) added a switch to apply miniIsolation 
2) Switched off the selfVeto in miniIsolation
3) adapted dxy and dz for inclusive leptons for electrons for PHYS14
Heppy: added miniIso to the list of variable for leptons

Added the histogram to the skimTrees.py script

MT2: switched to PHYS14 id electrons and miniIsolation for leptons plus various add on to the tree

### DIFF
--- a/PhysicsTools/Heppy/python/analyzers/objects/autophobj.py
+++ b/PhysicsTools/Heppy/python/analyzers/objects/autophobj.py
@@ -149,7 +149,7 @@ photonType = NTupleObjectType("gamma", baseObjectTypes = [ particleType ], varia
     NTupleVariable("chHadIso04",  lambda x : x.chargedHadronIso(), float, help="chargedHadronIsolation for photons (PAT method, deltaR = 0.4)"),
     NTupleVariable("neuHadIso",  lambda x : x.recoNeutralHadronIso(), float, help="neutralHadronIsolation for photons"),
     NTupleVariable("phIso",  lambda x : x.recoPhotonIso(), float, help="gammaIsolation for photons"),
-    NTupleVariable("mcMatchId",  lambda x : x.mcMatchId, int, mcOnly=True, help="Match to source from hard scatter (pdgId of heaviest particle in chain, 25 for H, 6 for t, 23/24 for W/Z), zero if non-prompt or fake"),
+    NTupleVariable("mcMatchId",  lambda x : x.mcMatchId if getattr(x,"mcMatchId",None) else 0. , int, mcOnly=True, help="Match to source from hard scatter (pdgId of heaviest particle in chain, 25 for H, 6 for t, 23/24 for W/Z), zero if non-prompt or fake"),
     NTupleVariable("mcPt",   lambda x : x.mcGamma.pt() if getattr(x,"mcGamma",None) else 0., mcOnly=True, help="p_{T} of associated gen photon"),
 ])
 
--- a/PhysicsTools/Heppy/python/analyzers/objects/autophobj.py
+++ b/PhysicsTools/Heppy/python/analyzers/objects/autophobj.py
@@ -53,6 +53,7 @@ leptonType = NTupleObjectType("lepton", baseObjectTypes = [ particleType ], vari
     # Isolations with the two radia
     NTupleVariable("relIso03",  lambda x : x.relIso03, help="PF Rel Iso, R=0.3, pile-up corrected"),
     NTupleVariable("relIso04",  lambda x : x.relIso04, help="PF Rel Iso, R=0.4, pile-up corrected"),
+    NTupleVariable("miniRelIso",  lambda x : x.miniRelIso if hasattr(x,'miniRelIso') else  -999, help="PF Rel miniRel, pile-up corrected"),
     # Charge flip rejection criteria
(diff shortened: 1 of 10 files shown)


gpetruc:
Hi Maria,

I have some issues with this pull request:
- miniIsolation: there's no need to add yet another parameter for miniIso given that `loose_muon_isoCut` and `loose_electron_isoCut` can already be used to cut on mini-iso (and indeed we could consider to deprecate out all the other isolation options, which lead to more parameters and complications for relIso, absIso, ...) 
- `PHYS14` is not a good name for the electron tightId choice (given that e.g. there's also a MVA for Phys14), so it should be something more like `Cuts_PHYS14_25ns_v1_ConvVeto`
- you've committed a lot of changes in the whitespace in the lepton analyzer (which one should try to avoid if possible, as it confuses reading the diffs), and one of those changes seems to be bad (`looseMuons = event.selectedLeptons[:]` is now within the loop on the muons, instead of without)
- I definitely don't want any hard-coded dxy-dz cuts and the `POG_Cuts_ID_PHYS14_25ns_v1_ConvVeto_Veto_full5x5` in the lepton analyzer code: please define an extra id instead (e.g. `POG_Cuts_ID_PHYS14_25ns_v1_ConvVetoDxyDz_Veto`)

mariadalfonso:
@gpetruc @emanueledimarco

As you requested:
1) I fixed the indent in the looseMuons
2) I adjusted the label for the  tightId  choise
3) I left on the self veto in miniIsolation for the  muons, and now it's off for all the options we have for the electrons.

Not done:
1) I would not deprecate the other isolation options for now.
I left the extra parameters loose_muon_isoCut and loose_electron_isoCut as good to avoid to everybody to switch to the miniIso. We can clean up later before migrating to 74X.
2) I added the POG_Cuts_ID_PHYS14_25ns_v1_ConvVetoDxyDz_Veto in the Lepton.py.
However I have problems in some very rare events with the dxy, dz and lostHits,  such that looks like are not applied.
(lost hits it's not applied properly neither in POG_Cuts_ID_PHYS14_25ns_v1_ConvVeto_Veto)
I could't understand why, so I left them hardcoded in the miniIsolation

mariadalfonso:
@gpetruc 

bug fixed line
https://github.com/CERN-PH-CMG/cmg-cmssw/blob/CMGTools-from-CMSSW_7_2_3/PhysicsTools/Heppy/python/physicsobjects/Electron.py#L105

Fixed the leptons convDxDy and removed the hard coded id from the leptonAnalyzer

PhysicsTools/Heppy/python/physicsobjects/Electron.py
```diff
@@ -43,7 +43,7 @@ def cutBasedId(self, wp, showerShapes="auto"):
             showerShapes = "full5x5"
             wp = wp.replace("_full5x5","")
         elif showerShapes == "auto":
-            if "POG_CSA14_25ns_v1" in wp or "POG_CSA14_50ns_v1" in wp or "POG_PHYS14_25ns_v1" in wp:
+            if "POG_CSA14_25ns_v1" in wp or "POG_CSA14_50ns_v1" in wp or "POG_PHYS14_25ns_v1" in wp or "POG_PHYS14_25ns_v1_ConvVeto" in wp or "POG_PHYS14_25ns_v1_ConvVetoDxyDz" in wp:
```
gpetruc, line 46: this change is not needed: `POG_PHYS14_25ns_v1` is contained in `POG_PHYS14_25ns_v1_ConvVeto` etc

PhysicsTools/Heppy/python/analyzers/objects/autophobj.py
```diff
@@ -53,6 +53,7 @@
     # Isolations with the two radia
     NTupleVariable("relIso03",  lambda x : x.relIso03, help="PF Rel Iso, R=0.3, pile-up corrected"),
     NTupleVariable("relIso04",  lambda x : x.relIso04, help="PF Rel Iso, R=0.4, pile-up corrected"),
+    NTupleVariable("miniRelIso",  lambda x : x.miniRelIso if hasattr(x,'miniRelIso') else  -999, help="PF Rel miniRel, pile-up corrected"),
```
gpetruc, line 56: You added it here, but it was already in `leptonTypeSusy`. If you think it should be here, then remove it from there to avoid duplicates.

gpetruc:
I think the miniIso can still be implemented more cleanly.
I add @arizzi @cbernet to the thread since eventually what we're discussing ends up in Heppy

**Options:**
1. if you want to apply a miniIsolation cut, just put in the cfg the cut, e.g. `loose_muon_isoCut = lambda muon : muon.miniRelIso < 0.4`
   - **pros:** doesn't add any new parameter in the cfg, needs _no_ changes in the lepton analyzer code, it's immediately clear what it does, and goes in the direction of allowing to deprecate the isolation cut values in favour of using isoCut in future (and thus remove some parameters and make the code even shorter than now)
   - **cons:** going back from the lambda to the number is not immediately possible 
2. add one boolean `useMiniIsolation`, if set to `True` then the  `loose_muon_relIso` and possible `loose_muon_absIso` will cut on mini-isolation instead of relIso(R=0.3), which can be done at the time the cut function is defined (without bringing the if inside of the lambda)
   - **pros:** you can still have the numbers out of the config, and the changes are still much less than what's now in this PR
   - **cons:** you still need to modify the code that defines the iso cut, you still add one more parameter, people need to look at two parameters at the same time (the cut value and the switch), and doesn't extend to any future alternative isolation
3. identical to above, but instead call the parameter `isolationDefiniton`, default to `relIso03`, then the cut function can just cut on `getattr(lepton, isolationDefiniton) < cut`:
   - **pros:** more scalable than 2, ang adds in support for relIso04 for free
   - **cons:** still less compact than 1

PhysicsTools/Heppy/python/analyzers/objects/autophobj.py
mariadalfonso, line 56: removed the miniRelIso from the leptonTypeSusy and kept the one in heppy

cbernet:
Hi,
It seems to me that solution 1 is the best one. It's clean and much more generic. For instance, such a solution will work if somebody comes up with a nano isolation (or whatever) in the future.

PR #338, opened 2015-04-07
H2TauTau: SVfit updates, added QCD to samples

rmanzoni:
Implement latest changes to SVfit by Christian: reduce 250000 --> 100000 calls to evaluate likelihood in MarkovChain integration and disable redundant debug printouts.
Aligned to https://github.com/veelken/SVfit_standalone/tree/svFit_2015Apr03

Add QCD samples to qcd.py and connector.py under phys14.

### DIFF
--- a/CMGTools/H2TauTau/python/proto/samples/phys14/connector.py
+++ b/CMGTools/H2TauTau/python/proto/samples/phys14/connector.py
@@ -8,6 +8,7 @@ from CMGTools.RootTools.utils.splitFactor                   import splitFactor
 from CMGTools.H2TauTau.proto.samples.phys14.higgs           import mc_higgs
 from CMGTools.H2TauTau.proto.samples.phys14.ewk             import mc_ewk
 from CMGTools.H2TauTau.proto.samples.phys14.diboson         import mc_diboson
+from CMGTools.H2TauTau.proto.samples.phys14.qcd             import mc_qcd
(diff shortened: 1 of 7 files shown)


steggema:
There are conflicts with e.g. https://github.com/CERN-PH-CMG/cmg-cmssw/blob/CMGTools-from-CMSSW_7_2_3/CMGTools/H2TauTau/cfgPython/mt/tauMu_2014_cfg.py#L22
I think we should go to the command-line solution. Can you update the PR?

The other thing is that I see several whitespace modifications in the SVFit code, e.g. here https://github.com/CERN-PH-CMG/cmg-cmssw/pull/338/files#diff-c5a38d45896330f05ea5078b941d2a00R25, but I think I don't see them in Christian's code. If you added them, I think it would be good to revert them as those would be reverted the next time we copy code from Christian's repository, adding unnecessary noise to the history.

rmanzoni:
The conflicts appeared after the merging this PR https://github.com/CERN-PH-CMG/cmg-cmssw/pull/329
Of course the command line solution is much better, I stuck to that as far as mt goes (et, tt, em need to be updated accordingly)

The whitespace thing has been done automatically by my editor. This can be fixed for sake of clarity, of course.

My proposal: discard and close this PR, and I'll submit a new one, with a cleaner history, in a few mins, ok for you?

rmanzoni:
Keep this PR alive if you like, I 'reset --hard' and recommitted what needed to.

PR #339, opened 2015-04-07
Cmg tools cleanup

mariadalfonso:
Removed duplicate files in CMGTools since are already in Heppy
Fixes in some Heppy files left over dependency from CMGTools.


gpetruc:
Fine for me. @steggema is it ok for H->tau tau as well?

steggema:
Two things:

1) We're currently using the VBFSimpleAnalyzer, and I don't like too much the idea of using something that's in heppy examples, but we may simply create and use a copy in our H->tau tau directory in a subsequent PR, so that's ok for me.

2) I see that the TriggerAnalyzer wouldn't even run with heppy since it uses the old-style syntax. I realise that this is the case for several other of the example analyzers, so my question would be if we want to fix it in this PR or in a later one. Let me know.

gpetruc:
Ciao,

I agree that using something that is in heppy examples is a bad idea.
I you think you can find some time in the next few days to do your copy of the VBF analyzer and possibly fix the other ones, then maybe we can hold this PR and you can start from it.
Otherwise we could merge this and then we fix them later.

mariadalfonso:
@steggema 
regarding 1) I prefer to avoid duplication in other HtauTau packages and just put your the VBFanalyzer in the heppy main area . If there are other cases like this, suggest to move to heppy common package as well.
regarding 2) in particular, I think it's worth to update at least the ones that you are using (and move to the main package if needed) maybe before next week meeting

steggema:
@mariadalfonso 
1) Maybe we can discuss this in next week's meeting: I don't think that we need analysis-specific code in heppy, but there may be different opinions on what's analysis-specific and what not. For me, it's a perfect use case to have the VBF analyzer in examples as well as in the tau-tau code, as example code should be copied and modified. We can also have it only in tau-tau as I don't know of anyone else who is using this, though I think it's a reasonable example of how to use the VBF class.
2) The one analyzer we're using are of course working, we'll just have to adapt the import statements.

My suggestion would be to merge as is, and we'll add the modifications including possible fixes of other heppy example analyzers (e.g. we should probably remove some of the redundant ones like the JetAnalyzer and MetAnalyzer) in a separate PR.

PR #343, opened 2015-04-09
Merge heppy post 342

gpetruc:
Merge the heppy branch in CMGTools after #342 and preserve backwards compatibility in Susy/TTH and MonoX subsystems (H->ZZ->4l will change in #334)

@steggema does anything need to be done for tau tau?


steggema:
Fine for tau-tau; we're currently not using any of the standard lepton analyzers.

PR #344, opened 2015-04-10
CMG: MetAnalyzer restructuring for multiple instances

schoef:
Continuing with METAnalyzer: Default behaviour should not be changed.
METAnalyzer can now be run more than once.

*) added switches for copyByValue (default=False. This is needed because shiftPt functionality is lost when 
copy-byvalue is done)
*) made input collection steerable (this is useful when running on private mAOD with additional MET collections)
*) made all MET collections class members, added collectionPostFix steerable for storing in event
*) interfaced with JetAnalyzer: Now the JEC and JER deltaMETs can be selected by aligning the postFixes appropriately
*) I think noPUMet should be removed but it is still used in multilep so I kept it. I made a switch, though.
One more issue with this: By default, it uses the same input as the default MET class, therefore, when the JEC corrections are turned on, I think they will be applied twice because a copy-by-value is assumed but not done.


schoef:
Hi,
any issues with this

gpetruc:
Ciao,

In 74X, we had problems with deepcopy, that we solved calling the pat::MET construtor directly 
https://github.com/cbernet/cmssw/commit/1d6a7afa64588691bf457a4fe9a4865fd56f32e3
(which required the fix https://github.com/cbernet/cmssw/commit/1d6a7afa64588691bf457a4fe9a4865fd56f32e3 , to be backported to 72X)

I was wondering it we want to do this already in 72X in order to ease further ports from 72X to 74X?

 Giovanni

schoef:
PR is updated.
It compiles.

PR #354, opened 2015-04-27
Updates for monoJet analysis

emanueledimarco:
several simple updates:
- add some first tree analysis code (ported from tth analysis)
- move the razor event topology variables from CMGTools to Heppy
- update samples for PHYS14 set
- put the triggers for monojet MC in a separate file included in the samples file and update the path list
- merge with the latest Heppy and CMGTools version in CMSSW_7_2_3


mariadalfonso:
@emanueledimarco 

Why you have all these commits Things should have been merged already.
https://github.com/CERN-PH-CMG/cmg-cmssw/pull/354/commits

Those seems are not only monoJet related

Also avoid to change files like this
https://github.com/CERN-PH-CMG/cmg-cmssw/pull/354/files#diff-78629d8889b265fe7ae9d972816817cfL98

emanueledimarco:
@mariadalfonso 
because I merged the updates before starting implementing new features. I think that this is safe: the list of modified files are only the ones that I modified, the other commits are already merged and there should not be any conflict. 

Regarding the white line removal in PhysicsTools/Heppy/python/analyzers/objects/METAnalyzer.py, I didn't realised it, but it since it is harmful, I would not do a girandola  of git commands to  revert it back :)

PhysicsTools/HeppyCore/scripts/heppy_batch.py
```diff
@@ -63,7 +63,7 @@ def batchScriptCERN( jobDir, remoteDir=''):
 do
    ff=`basename $f | cut -d . -f 1`
    #d=`echo $f | cut -d / -f 2`
-   gfal-mkdir -p {srm}
+   gfal-mkdir {srm}
```
gpetruc: I think the version with -p may be the more correct one ( @brucala please confirm )
if yes, can you please add a commit to put back the -p ?

steggema:
Concerning the comment by @mariadalfonso 
Are we sure we understand what's going on? Usually if you make a merge while developing, github will only display the merge commit (see e.g. the 2nd commit here https://github.com/CERN-PH-CMG/cmg-cmssw/pull/349/commits), but not all commits that were applied during the merge. Even though the file changes look clean, I would want to make sure that we don't import a parallel history.

gpetruc:
Hi @steggema 

I think what we're seeing is that we have duplicated commits because ofthen the heppy ones are cherry-picks or rebases of the CMGTools ones.
Thus, we in the merge we add a commit to the graph, even if t does not change anything

Giovanni

steggema:
Hi @gpetruc 

Ok I see - it's because of this merge https://github.com/emanueledimarco/cmg-cmssw/commit/2c756eecd94d53e8ebed79c38c6767f5cf35674f from the heppy branch, so we'll generate a duplicate history by means of what you write. 
In general, I would be in favour of trying to avoid a duplicate history if possible (since there should be no need to explicitly merge with the heppy branch), but it doesn't really matter in this case since we intend to drop the history anyway when going to 74X.

thanks, Jan

emanueledimarco:
Hi @steggema and @gpetruc - I noticed that, but I thought that it is harmless (apart that it seems a PR with hundreds of commits). How should I do to avoid this?

steggema:
Instead of merging from the heppy branch like in https://github.com/emanueledimarco/cmg-cmssw/commit/2c756eecd94d53e8ebed79c38c6767f5cf35674f, you can only merge from the CMGTools branch since that should have all heppy commits (in the sense of diffs) applied via rebasing. Was there any particular reason why you merged from the heppy branch?

emanueledimarco:
@steggema Because I also did some small change in heppy, and I was not sure that all the latest commits in heppy were already merged in the CMGTools branch. So I first merged that, and then CMGTools, and finally did the PR to CMGTools

steggema:
For example, here's how the history looks like after this PR is merged:
https://github.com/steggema/cmg-cmssw/commits/test_pr354/PhysicsTools/Heppy/python/analyzers/objects/METAnalyzer.py (the 2nd and the 5th are the same commits with different IDs)

Concerning the heppy branch, our strategy has been to rebase (aka cherry-pick) the commits from the heppy branch and apply them to the CMG branch, or vice versa, and not to merge. 

That's maybe something to rethink for the future, but the merge only works one way as opposed to the rebase strategy.

cbernet:
Hi all, 

I also don’t like that too much… I’m afraid it’s going to be difficult to maintain… 
Cheers, 

Colin

gpetruc:
Hi,

git merge handles correctly the cherry-picked commits, the only issue is
that we should do more often the merge of heppy into cmgtools (ideally
every time we merge a heppy PR) so that we avoid having spurious commits in
the PR.
Maybe we can decide that we always do _silently_ the merge into CMGTools
every time we merge a PR into heppy (i.e. from the command line), a bit
like e.g. every time a PR is merged in 74X then the 74X can be remerged in
75X.

```
Giovanni
```

PhysicsTools/HeppyCore/scripts/heppy_batch.py
gpetruc: ping .... can we fix this three-character difference so that we can merge the PR?

gpetruc:
Also, we need to have heppy mergeable cleanly in CMGTools, otherwise we have to rebase each and every commit whenever somebody else contributes to heppy outside cmgtools, which makes no sense.

PR #355, opened 2015-04-28
Parametric batch submission at IC, toggle for LHE reading and always clean jets from photons

aelwood:
- Implemented parametric batch submission at IC. This is to circumvent the hard 2000 individual job limit. A file with all the job directories is saved and the jobs are submitted separately in a parametric way.
- The non lazy LHE extra information event reading was found to significantly increase the timing of the susyParameterScanAnalyzer (and GenAnalyzer) (increasing the fraction of time they took by ~10%) . Added a toggle to switch it off for the susyParameterScanAnalyzer). 
- Added an option to define the cleanJets as ones that are cleaned from photons as well.

### DIFF
--- a/PhysicsTools/HeppyCore/scripts/heppy_batch.py
+++ b/PhysicsTools/HeppyCore/scripts/heppy_batch.py
@@ -211,7 +211,7 @@ echo 'running'
 python {cmssw}/src/PhysicsTools/HeppyCore/python/framework/looper.py pycfg.py config.pck
 echo
 echo 'sending the job directory back'
-cp -r Loop/* ./ && rm -r Loop
+mv Loop/* ./ && rm -r Loop
 """.format(jobdir = jobDir,cmssw = cmssw_release)
    return script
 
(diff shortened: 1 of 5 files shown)


PhysicsTools/HeppyCore/scripts/heppy_batch.py
```diff
@@ -211,7 +211,7 @@ def batchScriptIC(jobDir):
 python {cmssw}/src/PhysicsTools/HeppyCore/python/framework/looper.py pycfg.py config.pck
 echo
 echo 'sending the job directory back'
-mv Loop/* ./
+cp -r Loop/* ./ && rm -r Loop
```
gpetruc: why this change?
usually move is cheaper than copy + delete (and doesn't use up twice the space on disk or afs while it is performed)

aelwood: The "Loop" directory still existed, which meant when resubmitting "Loop_1" would be made instead. I can change this to a mv then rm instead of a cp and rm.

gpetruc:
Hi @aelwood , can you make the corresponding Heppy PR?

aelwood:
@gpetruc done https://github.com/CERN-PH-CMG/cmg-cmssw/pull/359

PR #358, opened 2015-05-03
Heppy: MT2 from CMGtools to heppy + small additions to METanalyzer

mariadalfonso:


gpetruc:
waiting for fix as in #357 
Please always put links between Heppy PRs and CMGTools PRs, it just takes a hash sign and 3 digits in the comment box...

PR #365, opened 2015-05-04
fix a crash in the Photon fiducial regions

emanueledimarco:
the baricenter of the cluster of a photon can be at |eta|>2.5 (EE arrives at eta=3.0).
The PhotonAnalyzer was crashing because of that sometimes.


emanueledimarco:
After discussion with @pandolf and Giovanni Zevi, moved to the variable used in the cut based photon ID.

aelwood:
Is there a reason this isn't merged into the CMGTools-from-CMSSW_7_2_3 branch, or is it ok for me to move it across for our tree production?

PR #373, opened 2015-05-11
Updates for MonoJet signal region and Z and W control regions

emanueledimarco:
- Several changes in the selection chain for the signal region and the control regions
- plot some std variables for all the regions
- add friend trees for event variables
- merge the latest CMGVersion and update all the analyzers to this


gpetruc:
Eventually we should refactor most of prepareEventVariablesFriendTree somewhere else, since eventually the only thing you need different is the definition of the actual modules to run.

emanueledimarco:
@gpetruc I agree. I took most of the tools from TTHAnalysis/macro indeed. Maybe we can do this when I move to 740 the MonoXAnalysis

PR #374, opened 2015-05-18
Porting MonoXAnalysis to 7_4_0

emanueledimarco:
- There are many files removed, but for some reason I mistakenly committed the output of a GEN production in 7_2_3 that went to GIT. They are the majority of the 840 files.
- do not changed the samples lists (still PHYS14 samples)


emanueledimarco:
This CMGTools branch also contain the porting of the event topology analysers to heppy in the 7_4_0 branch, corresponding to the PR #375

emanueledimarco:
I will do the porting to the new 7_4_3 branch.

PR #375, opened 2015-05-18
heppy: add event topology analyzers in heppy

emanueledimarco:
This is already done in 7_2_3. Adding back the directory with the Razor, alphaT and MT2 analysers also in the 7_4_0 branch.


emanueledimarco:
Not needed any more, since everything is in the 742 branch

PR #386, opened 2015-05-27
stuff for soft-opposite-sign leptons analysis

cbotta:
@rcastello @gpetruc @ikesisog-atCMS 
This is for soft-opposite-sign lepton analysis.
I removed the old cfgs, and add a new one which makes use of the same treeProducer of the Multilepton analysis.

cristina


gpetruc:
Hi Cristina,

You seem to have deleted also the `run_susySingleSoftlepton_cfg.py` which was used for the single-lepton analysis, not the SOS one. I suppose that one should not be deleted (unless @safarzad or @mstoye say it's obsolete)

   Giovanni

cbotta:
Hi,

it was intentional, 
I discussed that with @mstoye and @kirschen this morning

cristina

PR #388, opened 2015-06-02
74X PR which follows changes in 72X PR #382

mangano:
74X port of  PR #382


mangano:
Added to this PR also the changes from Emanuele as in #394, resolving conflicts

PR #389, opened 2015-06-02
Heppy PR following CMG PR #388

mangano:
Heppy version of #388
- add configurable jetEta cut to cleanJetsAll collection
  PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
- modification for photon sidebands (from F.Pandolfi):
  PhysicsTools/Heppy/python/analyzers/objects/PhotonAnalyzer.py
  PhysicsTools/Heppy/python/physicsobjects/Photon.py

PR #394, opened 2015-06-03
Heppy: Fixes for E/gamma Electron ID and Photon ID in 74X

emanueledimarco:
This PR includes two commits:
- The https://github.com/CERN-PH-CMG/cmg-cmssw/commit/9cc763e37637e10ffd469b95df7a8450469ea544 adds the V2 Cut Based electron ID, which is the new optimisation recommended by the POG, see: https://twiki.cern.ch/twiki/bin/viewauth/CMS/CutBasedElectronIdentificationRun2#Working_points_for_PHYS14_sample 
  In order not to introduce disruptions in who is using V1 version, the  new one is called as before, but with V2 label
- The https://github.com/CERN-PH-CMG/cmg-cmssw/commit/a09977ad8e937529e054eda765bec4f5fab13e68 contains a change for I think a bug in the scaling of the isolation for neutral hadrons. The offset and slope of the correction in https://twiki.cern.ch/twiki/bin/view/CMS/CutBasedPhotonIdentificationRun2#Working_points_for_PHYS14_sample were swapped. 
  The second change is that now it seems that in the EB only for NH isolation, the relationship is no more linear, but exponential. I implemented this now. This replaces the old one (no V1 or V2 here)


emanueledimarco:
updated the branch with a new addition:
- The https://github.com/emanueledimarco/cmg-cmssw/commit/bc06fd19683ce36dfef443e3d9cd617bdf5766ff commit includes the usage of the rho corrections and applies them to the photon WPs. This is recommended by the e/gamma POG (otherwise the cuts are not consistent with the variables used to derive them)

emanueledimarco:
I added the functionality needed by MT2 analysis. If the string like "POG_PHYS14_25ns_Loose_looseSieie_NoIso" is given to the PhotonAnalyzer, then isolation is not used at all, moreover:
- it uses looser sigma(ieta ieta) cuts [EB,EE] = [0.015,0.035] instead of the nominal ID cuts to select the good photon candidates
- it sets the variable gamma.idCutBased as the output of the nominal ID (in this case "POG_PHYS14_25ns_Loose")

This works for any of the cut based WPs defined in https://github.com/emanueledimarco/cmg-cmssw/commit/99768fe8a9397f700ab17f38331b055563b5eeb2#diff-05d41322b9e8e0dccfe521d6e1d1131c

emanueledimarco:
All my changes (fix of the photon isolation, fix of the photon ID cuts, implementation of the sidebands for Pandolfi) are now merged in the #389 from Boris. Then I am closing this.

PR #398, opened 2015-06-08
CMG: fixed edmIntegrityCheck (xrdcp no force option anymore)

cbernet:


gpetruc:
uhm, perhaps the `-force` should become `--force` , instead of removing it completely?

cbernet:
Indeed, the script works with --force. Note that the xrdcp error message was saying that -force was not supported anymore, and that it should be replaced by --retry. Tried that, didn't work. That's why I had chosen to remove this option algtogether. But here is an update anyway.

PR #405, opened 2015-06-16
Synch. 74x revised

emanueledimarco:
- This version has the fixed version of the tauID and has the new tables of the synch cut flow as in the monojet twiki. 
- The event variables in the friend trees are better defined


emanueledimarco:
This pull request includes #407

PR #408, opened 2015-06-23
Add CountLHE histogram to include the sum of weights for the total nu…

jmvizan:
…mber of processed events (regardless they pass the skim or not), corresponding to the different LHE weights stored in the LHEEventProduct branch (the bin center corresponds to the different LHE ids and the content to the corresponding sum of weights). Modification already present in the previous branch for CMSSW_7_2_3)


gpetruc:
will merge #415 instead

PR #413, opened 2015-07-01
Heppy: Fixed label for LHE weights, plus added original LHE weight

mmasciov:

PR #414, opened 2015-07-01
Added LHE weights info, plus added histo with sum of gen weights. Related to Heppy PR #413

mmasciov:
Related to Heppy PR #413


mmasciov:
Closing and re-opening clean PR

PR #416, opened 2015-07-01
Added LHE weights info, plus added histo with sum of gen weights. Related to Heppy PR #415

mmasciov:
Don't understand why previous commits from other contributors appear in my PR? 
However, the list of changed files (3) is correct.


gpetruc:
@jmvizan 
I think this includes already what's in #408 so maybe merging just this is enough?

jmvizan:
Yes, for what I see all the information I added in #408 is also included in this pull request

PR #424, opened 2015-07-03
Sample move

mariadalfonso:


mariadalfonso:
@emanueledimarco 
I moved all the samples and triggers files in 
../../RootTools/python/samples/samples_13TeV_74X.py

PR #425, opened 2015-07-06
Heppy lep updates

gpetruc:
- Support for FSR-aware isolation
- Add heppy_report.py (produce a JSON of all processed lumisections)


gpetruc:
will merge in `heppy_74X`, not here.

PR #420, opened 2015-07-02
Merge of Colin's 7_4_5 into CMGTools

gpetruc:
Like #419 but with a few extra updates needed on the CMGTools side:
- be26bc4 : susyCore_modules_cff: Change default JECs from PHYS14_25_V2 to Summer15_V5_MC
- e830f14 : Adapt cfgs to the upstream change framework.heppy -> framework.heppy_loop

PR #419, opened 2015-07-02
Update Heppy branch from Colin's 7_4_5

gpetruc:
The diff below is unreadable because it includes multiple updates of CMSSW versions, but you can get a comparison of (cmg `heppy_743` + `CMSSW_7_4_5`) vs Colin's `heppy_7_4_5` from https://github.com/gpetruc/cmg-cmssw/compare/cmg-heppy_743-merged-in-CMSSW_7_4_5...gpetruc:cbernet-heppy_7_4_5

The only commit added on top of @cbernet 's branch is 4eec8fe663e0d1b52d09b670c55306641ad4ced8

PR #428, opened 2015-07-09
Update to 747 and Colin's heppy745 

gpetruc:
**Important notice 1:** if you use CMSSW_7_4_7, you need also `DataFormats/FWLite` in the sparse checkout; this PR already includes the fix https://github.com/cms-sw/cmssw/pull/10160 to a bug there.

**Important notice 2:** this PR changes the SUSY jetAnalzer defaults to apply `Summer15_V5_p6_MC` jet energy corrections on data and MC (the past default, Phys14 v2, was clearly a leftover from 72X)

There seems to be sometimes crashes when running multithreaded in the new ipython, which are still being investigated (see #419 #420). Because of this, for the moment, `heppy` will run python, and `iheppy` will run ipython.


gpetruc:
@mariadalfonso can you make another test on top of 7_4_6_patch6?

then, I will merge into a _new_ CMGTools-from-CMSSW_7_4_6_patch6 branch.

mariadalfonso:
@gpetruc 
I tested the PR428 on top of the 743 branch and no changes http://dalfonso.web.cern.ch/dalfonso/validationPR428plots/
For me we can move heppy and cmg branches to the 7_4_6_patch6 (in synch with the release used for the data).

PR #435, opened 2015-07-12
heppy: fix JetAnalyzer copy by value 

schoef:


gpetruc:
will merge in `heppy_74X`, not here.

PR #437, opened 2015-07-12
H->ZZ->4l 74X, and small heppy and python samples updates

gpetruc:
- Port of H &rarr; ZZ subsystem
- Add updated Phys14 MVA Ele Id training (for H&rarr;ZZ)
- Support for caching DAS files for a given run range (important for running on PromptReco, otherwise the framework will not re-query DAS and will not discover new files)
- Fix the application of `Summer15_V5_p6_MC` corrections on data
- Some minor heppy bugfixes


gpetruc:
@mariadalfonso: I'm done with updates with the updates to this PR.
It should be transparent to other susy except for the setting by default of Summer15_V5_p6_MC also for Data (it was supposed to be done already in #428 but that configuration was not working on data), but it would be good if you can confirm it.

PR #438, opened 2015-07-13
Changes to central repo from RA1 port to 747

aelwood:
- Added option to do relative isolation in the isoTrackAnalyzer
- Removed redundant AlphaT analyzer in TTHAnalysis
- Allowed the use individual pile up correction for different leptons when miniIsolation is applied (previously could only choose one type for all objects with miniIsolation)
- In the config choose 
    miniIsolationPUCorr = None
  and it'll fall back to the pu correction defined for the individual objects


mariadalfonso:
@aelwood . Can you make a PR of the heppy files that contains #438 and #440 in the heppy_74X branch ?

aelwood:
@mariadalfonso , made here https://github.com/CERN-PH-CMG/cmg-cmssw/pull/449

PR #441, opened 2015-07-14
Updates from DESY/ACDV

artlbv:
ACDV related:
- created SUSYAnalysis dir in CMGTools
- updates in plotter/susy-1lep
- updated nafbatch script
- separate trigger file in samples

General:
- Create Component for files at DESY
- Create Component from directory
- extraLabel option in mcPlots


mariadalfonso:
@artlbv 
We should discourage to have extra packages in the central release just for plotting CMGTools/SUSYAnalysis/python/tools
Why those cannot be added inside the TTH  CMGTools/TTHAnalysis/ ?
those already contains All the SUSY analysis

artlbv:
@mariadalfonso hi, I just got inspired by https://github.com/CERN-PH-CMG/cmg-cmssw/tree/CMGTools-from-CMSSW_7_4_7/CMGTools/MonoXAnalysis/macros
since we have a separate macro to prepare friend trees, I didn't want to conflict with the default TTH one.
So it's mainly to separate the friend tree production part, but also to keep separate DESY sample + trigger files.
What would be an alternate location for that?

mariadalfonso:
We can create a directories  insides
CMGTools/TTHAnalysis/python/tools
"macros1l"

@cbotta can you also move the specific SS analysis in a direcrtory  "macrosSS" os similar
We already has such a scheme  in CMGTools/TTHAnalysis/python/plotter/susy-1lep/

artlbv:
Will fix this in another PR.

PR #442, opened 2015-07-15
Implementation of egamma-like footprint removal in photon isolation

emanueledimarco:
Synchronised with e/gamma full framework version on Spring15 MC sample.

PR #443, opened 2015-07-15
Migration of MonoXAnalysis to CMSSW_747

emanueledimarco:
- Use the fixed photon isolations in the trees. This PR indeed includes the heppy one: #442 
- Many small changes to synch the signal and control regions
- add the samples for Run2015B data (for the time being in the MonoXAnalysis samples file, will remove it when they will be added in the RootTools)
- add Z->ee control sample


emanueledimarco:
@gpetruc @mariadalfonso  I have merged my heppy_74X branch here and the up-to-date version of the CMGTools branch.

PR #450, opened 2015-07-20
Heppy: added option in JSONAnalyzer to count lumis based on luminosity...

brucala:
…  blocks instead of events


brucala:
sorry, i didn't validate on MC and forgot to add a protection for cases where there is no JSON. Will update the commit within the next 10 minutes

PR #467, opened 2015-08-06
Heppy: Isolation calculation for several cone sizes and fluctuation of event energy density

peruzzim:
Related to CMGTools PR #466


mmasciov:
Hi,
If I'm not mistaken, this change would add 10 extra floats per lepton in the basic lepton type, even if the isolation scan calculation is switched off.
Since not all of the users will need these 10 extra floats per lepton, wouldn't it be better to add them to the extra lepton type, or in alternative create even a different lepton type?

Mario

peruzzim:
@mmasciov thanks for spotting this, it was not meant to be in the PR. I have removed it, it will be included in the lepton content only by those who need it.

PR #469, opened 2015-08-07
More 25ns samples added and fixes to the x-sections

emanueledimarco:
- For CMGTools/RootTools:
  - add more 25 ns HT-binned samples (Z->inv, QCD HT, etc.)
  - fix some x-sections, putting the ones read in McM and suggested in https://twiki.cern.ch/twiki/bin/view/CMS/SummaryTable1G25ns
- For MonoX: 
  - update the list of samples to be run for 25ns MC


mariadalfonso:
@emanueledimarco 
The data sample are here RootTools/python/samples/samples_13TeV_DATA2015.py
So there is no need to have some extra data components in samples_monojet_13TeV_74X.py
Merging now, just for next time.

PR #470, opened 2015-08-08
add METnoHF and Type1Correction into the MT2tree

mariadalfonso:
This PR shows how to insert the METnoHF and the Tye1Correction into the susyTree
It use what is done in #468 + need to do the following below

echo /CMGTools/ObjectStudies/ >> .git/info/sparse-checkout
echo /PhysicsTools/PatUtils/ >> .git/info/sparse-checkout
echo /JetMETCorrections/Type1MET/ >> .git/info/sparse-checkout
echo /PhysicsTools/PatAlgos/ >> .git/info/sparse-checkout

-----> commit d375918 contains the  <----- 
git cms-merge-topic -u cms-met:METCorUnc74XSquashed .
already merged in the release

Changes:
1) added the V4 JEC corrections as default for both DATA and CM
2) residuals JEC switched on by default in susyCore

other nice general features:
1) add protection to genMET in case is not there see commit 9936c77
2) put heavy genMatching behyond a flag do_mc_match see commit ff626b2


gpetruc:
Hi @mariadalfonso , @mmarionncern: can you take what's in `git cms-merge-topic -u cms-met:METCorUnc74X`, squash it into a single commit (so that we can revert it easily when we update the CMSSW release after https://github.com/cms-sw/cmssw/pull/10443 is merged), and push it to this PR?
Thanks

mmarionncern:
@gpetruc , the changes of the 74X recipe will be directly in #10443 as the changes will differ slightly due to the change of format and the need that #10443 has to work on both AOD and mAOD.

gpetruc:
Hi @mmarionncern: my understanding from @mariadalfonso is that the recipe in this PR is to run whaveter possible in the current CMSSW releases without the dataformat changes that will go in a future CMSSW, and thus in terms of C++ are different from what's in 10443 and if we merge them we will then have conflicts when we will update to the new CMSSW + 10443.
However, as is the recipe in this PR is bad since it requires the user to merge things in his repository that are not to be merged in CMGTools, making development on top of the recipe impossible.
That's the reason for my request to compress all the extra changes in a way that they can be merged now, so that development can be done on top of it, and reverted later, so that we can update to the new CMSSW.

mmarionncern:
@gpetruc ,  Ok i understand better what you need, I will prepare that. However I cannot guarantee the recipe is now in its final state.

mmarionncern:
Hi @gpetruc , the squashed branch METCorUnc74XSquashed is available and can be used in a way similar to the standard recipe :
git cms-merge-topic -u cms-met:METCorUnc74XSquashed
N.B. : the latest JEC and corresponding uncertainties are plugged, but are still under test by the JetMET group

mariadalfonso:
@gpetruc This PR is ready and to be merged, go ahed if you are happy

gpetruc:
Ok, merged and updated the sparse checkout under `/afs/cern.ch/user/c/cmgtools/public/sparse-checkout_74X_heppy`

PR #472, opened 2015-08-10
Fixing JEC naming convention in JetRecalibrator.py to automatically account for centrally provided files

brucala:
Relative to Heppy PR #471

Note: jec global tags should be updated in the configuration files when necessary. See susyCore_modules_cff.py as an example.


gpetruc:
Ciao,
Are we sure that all present and future MC calibrations will end with  "MC" and the ones of data with "DATA" ? 
In the past, text files were called like the global tags (which IMHO is a good idea once things are in a GT), and those didn't end with DATA or MC

pandolf:
what i remember is that each GT has a set of MC and a set of DATA corrections.

maria can you confirm that this is still the case?

f

On 12 Aug 2015, at 12:21 , Giovanni Petrucciani <notifications@github.com<mailto:notifications@github.com>> wrote:

Ciao,
Are we sure that all present and future MC calibrations will end with "MC" and the ones of data with "DATA" ?
In the past, text files were called like the global tags (which IMHO is a good idea once things are in a GT), and those didn't end with DATA or MC

schoef:
We will know when the db files come out this evening.
There will likely be two files for data and MC and the JetRecalibrator 
should adjust to that.

brucala:
I had a look at the JEC payloads linked to this evening's meeting, and it looks like a MC and DATA version will be provided for all corrections (they can be a copy of each other). 
So it seems that this PR is in fact unnecessary at the end.

mariadalfonso:
@ closing this PR, since we do not need nay modification for the V4 JEC set

PR #473, opened 2015-08-11
crab tools update, compatible with 74X, avoid potential das querys

brucala:


gpetruc:
I don't see how this affects DAS queries.
Can you explain?
Thanks
Giovanni

brucala:
the line:
rm ~/.cmgdataset/CMS_PromptReco_
removes some datasets from the cahe, and if one tries later to create those components a das query would be called.
I don't think this was the problem on the current HN, cause this is for _PromptReco_ datasets, while the affected DAS query seem to come from QCD_HT samples.

gpetruc:
Ok, I had missed it at the first read.
Thanks

PR #479, opened 2015-08-16
DiJet codes for sbottom and susy monojet analyses

safarzad:
add codes for sbottom and susy monojet analyses


safarzad:
Forgot Labels which are local,susy/ttH

PR #480, opened 2015-08-17
added ntupler for WA7 measurement

mstoye:


gpetruc:
Why it is labelled as WMass but goes into TTHAnalysis?

mariadalfonso:
@gpetruc I mislabeled since this is an EWK measurement.

PR #481, opened 2015-08-17
updated trigger file for edge analysis

mdunser:


mariadalfonso:
@mdunser 
There is no need to duplicate the same triggers for all the analysis i.e. ht path are in common. Pick for example triggers_PFHTX00 [now we have triggers_pfht900, triggers_ht900, triggers_HT900] and then adjust the files that use them like the https://github.com/CMSEdge/cmg-cmssw/blob/CMGTools-from-CMSSW_7_4_7/CMGTools/TTHAnalysis/cfg/run_susyMT2_cfg.py#L199

gpetruc:
@mariadalfonso @mdunser : what fix are we waiting? and when is it supposed to arrive?

mariadalfonso:
@gpetruc @mdunser I will merge this one, and slim in a future PR the triggers list duplication (this is mainly on the aestetics)

PR #483, opened 2015-08-18
Heppy: Several small heppy updates from #482

steggema:


PhysicsTools/Heppy/python/physicsobjects/Lepton.py
```diff
@@ -44,6 +44,23 @@ def relIso(self,dBetaFactor=0, allCharged=0):
...
         return rel
 
+    def absIsoR(self, R=0.3, dBetaFactor=0, allCharged=False):
+        if dBetaFactor>0 and self.puChargedHadronIsoR(R)<0:
+            raise ValueError('If you want to use dbeta corrections, you must make sure that the pu charged hadron iso is available. This should never happen') 
+        neutralIso = self.neutralHadronIsoR(R) + self.photonIsoR(R)
+        #Recover FSR
+        if hasattr(self, 'fsrPhotons'):
+            for gamma in self.fsrPhotons:
+                neutralIso = neutralIso - gamma.pt()
```
gpetruc: For FSR, now we implement it in Electron.py and Muon.py since it needs to take into account the vetos in the photon isolation that are different for the different objects.
If you're not using FSR then maybe it's better if you remove the code from here, while if you use FSR then maybe it's better to use the absIsoWithFSR methods (which also supports rhoArea corrections instead of deltaBeta)

steggema: Ok agreed in principal - we're not using FSR. Should I then also remove it from the "absIso" function in Lepton.py? (making it backward incompatible in that respect)

Related: We can in principal also avoid the code duplication I inserted by removing the current "absIso" and renaming the new "absIsoR" to "absIso" at the cost of having a few more function calls.

gpetruc: Yes, I'd be in favour of dropping the old absIso and just have your new absIsoR.
Nobody is using the old absIso anywhere, at least from what github search reveals.

steggema:
Ok, updated both here and in the CMG branch. 

On something related: I think there is also some confusion with absEffAreaIso in the electron and muon classes vs the absIsoFromEA method in the lepton class. I would just remove the "absEffAreaIso" methods in the electron and muon classes, and then probably also remove the FSR removal in absIsoFromEA in the lepton class in line with this PR + modify the "relEffAreaIso" method in the lepton class.

https://github.com/CERN-PH-CMG/cmg-cmssw/blob/CMGTools-from-CMSSW_7_4_7/PhysicsTools/Heppy/python/physicsobjects/Lepton.py#L14
https://github.com/CERN-PH-CMG/cmg-cmssw/blob/CMGTools-from-CMSSW_7_4_7/PhysicsTools/Heppy/python/physicsobjects/Muon.py#L53
https://github.com/CERN-PH-CMG/cmg-cmssw/blob/CMGTools-from-CMSSW_7_4_7/PhysicsTools/Heppy/python/physicsobjects/Electron.py#L143

gpetruc:
Thanks,

yes, also those seem to be obsolete

Giovanni

steggema:
Ok, both this and #482 are updated in line with what's written above. 

Also to note that I put in another small fix (removal of non-used configuration attribute in the TauAnalyzer).

PR #485, opened 2015-08-19
add lheHT (genHT) as event variable (only for 1l-analysis so far), ta…

kirschen:
…ken from heppy LHEAnalyzer (needs to be run for this to work out)

Useful to e.g. combine inclusive TTBar samples with high HT-bin samples recently produced.


kirschen:
included commit from heppy PR #488 and edited treeproducer to write it into trees

PR #486, opened 2015-08-20
Lep-aware-JEC ptratio + tkJets + HBHE noise filter update + misc framework updates

peruzzim:
Summary of changes:
- added possibility of adding variables to object types at runtime (Heppy)
- added possibility of prefetching input files for preprocessor (Heppy)
- avoid overwriting event data members and lepton-jet association by analyzer modules (jet and MET) (Heppy)
- patch of DataFormats/FWLite to avoid crashes when preprocessing empty files
- updated configuration of HBHE noise filter with 50 and 25 ns cuts + added HBHE iso-based noise filter
- added lepton-aware-JEC calculation of the lepton ptratio variable
- added uncorrected ak4PFchs jets built with charged PF candidates from PV only
- fixed naming of metNoHF analyzer to avoid conflict with default MET analyzer
- configuration update for multilepton cfg

@gpetruc @cbotta

### DIFF
--- a/CMGTools/TTHAnalysis/cfg/run_susyMultilepton_cfg.py
+++ b/CMGTools/TTHAnalysis/cfg/run_susyMultilepton_cfg.py
@@ -535,7 +535,8 @@ from CMGTools.TTHAnalysis.tools.EOSEventsWithDownload import EOSEventsWithDownlo
 event_class = EOSEventsWithDownload if not preprocessor else Events
 EOSEventsWithDownload.aggressive = 2 # always fetch if running on Wigner
 if getHeppyOption("nofetch"):
-    event_class = Events 
+    event_class = Events
+    if preprocessor: preprocessor.prefetch = False
 config = cfg.Config( components = selectedComponents,
                      sequence = sequence,
                      services = outputService, 


mariadalfonso:
@gpetruc I tested marco's modifications on pre-processor / corMETMiniAOD_cfgCreator / METAnalyzer and are ok from my side.

CMGTools/TTHAnalysis/cfg/run_susyMultilepton_cfg.py
```diff
@@ -428,7 +532,7 @@
 # the following is declared in case this cfg is used in input to the heppy.py script
 from PhysicsTools.HeppyCore.framework.eventsfwlite import Events
 from CMGTools.TTHAnalysis.tools.EOSEventsWithDownload import EOSEventsWithDownload
-event_class = EOSEventsWithDownload
+event_class = EOSEventsWithDownload if not preprocessor else Events
 EOSEventsWithDownload.aggressive = 2 # always fetch if running on Wigner
 if getHeppyOption("nofetch"):
     event_class = Events
```
gpetruc: please add under `if getHeppyOption("nofetch"):` also a line `if preprocessor: preprocessor.prefetch = False`

gpetruc:
@peruzzim I don't like that now the multilep config requires a manual modification to switch between MC and Data, which breaks the validation scripts. Probably we should modify the CmsswPreprocessor so that it can also take two different cfgs one for MC and one for Data, and build both, since in any case when executed on a given component the preprocessor can tell whether it's MC or Data from `comp.isMC`.
We can discuss it tomorrow when we're both back at the office.

peruzzim:
Added fine splitting when running with preprocessor, from @gpetruc - to be more thoroughly tested before merging

gpetruc:
I'll take care of solving the merge conflicts in the preprocessor

PR #488, opened 2015-08-21
Heppy: add event.lheHTIncoming which restricts HT computation to particles t…

kirschen:
…hat have status<0 mothers (LHE)

needed to be able to use HT-binned TTBar-samples which exclude TTBar system from HT-computation


mariadalfonso:
@kirschen we need also the commit in the CMGbranch. 
For example add the commit into the #485 since is on the same topic

kirschen:
Hi Maria,
thanks for pointing that out! I just updated #485.
Cheers,
Henning

PR #494, opened 2015-08-25
Trigger matching analyzer and storage of several JEC levels

peruzzim:
- Heppy: added possibility to store the different JEC levels in the jet object
- Heppy: added possibility to add sub-objects to collection types at runtime
- Heppy: new module for matching of physics object to trigger objects
- minor configuration updates and features

### DIFF
--- a/PhysicsTools/Heppy/python/physicsutils/JetReCalibrator.py
+++ b/PhysicsTools/Heppy/python/physicsutils/JetReCalibrator.py
@@ -4,7 +4,7 @@ from math import *
 from PhysicsTools.HeppyCore.utils.deltar import *
 
 class JetReCalibrator:
-    def __init__(self,globalTag,jetFlavour,doResidualJECs,jecPath,upToLevel=3):
+    def __init__(self,globalTag,jetFlavour,doResidualJECs,jecPath,upToLevel=3,calculateSeparateCorrections=False):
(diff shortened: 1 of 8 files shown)


PhysicsTools/Heppy/python/physicsutils/JetReCalibrator.py
```diff
@@ -40,15 +57,20 @@ def correctAll(self,jets,rho,delta=0,metShift=[0,0]):
             print "Warning: %d out of %d jets flagged bad by JEC." % (len(badJets), len(jets))
         for bj in badJets:
             jets.remove(bj)
+        for j in jets:
+            for sepcorr in self.separateJetCorrectors.keys():
+                setattr(j,"CorrFactor_"+sepcorr,self.getCorrection(j,rho,0,[0,0],corrector=self.separateJetCorrectors[sepcorr]))
 
-    def getCorrection(self,jet,rho,delta=0,metShift=[0,0]):
+    def getCorrection(self,jet,rho,delta=0,metShift=[0,0],corrector=None):
         """Calculates the correction factor of a jet without modifying it
```
peruzzim, line 65: It is not true that this function does not modify the jet. If delta!=0, the attribute "jetEnergyCorrUncertainty" of the jet is set. The additional JEC levels can only be called with delta==0 to avoid overwriting jet attributes calculated with the standard corrector.

PR #501, opened 2015-08-28
Implement the Spring15 cut-based electron IDs

emanueledimarco:
This implements the WPs of the cut-based electron ID for both 50ns and 25ns. Respectively:
- Spring15_50ns_v1
- Spring15_25ns_v1
  It also includes the effective areas for DR=0.3 recalculated in Spring15 samples. 50ns is with the Run1 schema (flat PU target), 25ns is with the Run2 schema (using 68% contours to make efficiency flat).

Note: since the PF-isolation is recommended for DR=0.3, and there are not EAs calculated for DR=0.4, I have put EAs for DR=0.4 equal to 0, so that the code won't crash if requested, but it won't correct for PU at all.

MVA electron IDs for Spring15 are not yet implemented (even if 25ns non-triggering version is available from e/gamma).


emanueledimarco:
I modified few cuts for the 50 ns Spring15 electron ID cut based, moving to the V2 presented Friday here:
https://indico.cern.ch/event/369245/contribution/2/attachments/1153005/1655984/Rami_eleCB_ID_50ns_V2.pdf

PR #502, opened 2015-08-28
Use the Spring15 electron IDs and align to updated CMGTools

emanueledimarco:
- This includes the heppy PR #501. 
  It uses the Spring15 cut based IDs for the electron ID for monojet and uses again rho-area to correct for PU with the newest EAs.
- Then, sparse updates to align with the changes in CMGTools.


peruzzim:
A V2 version of the electron cut-based ID for Spring15 50 ns is going to be released by EGM, maybe it would be useful to include it in this PR already.

emanueledimarco:
I merged the #501 here to get the V2 cut based electron ID for 50 ns.

PR #512, opened 2015-09-16
Bugfix: usage of jet members not calculated by default in ntupleTypes

peruzzim:
Necessary to avoid crash when running without separate storage of different JEC levels.


gpetruc:
Maria, if it works for you can you merge?
Thanks 
Giovanni

PR #515, opened 2015-09-19
create lists of MiniAOD files with data in certified runs

TaiSakuma:
This PR addresses one of the issues raised in cmg-cmgtools@cern.ch.

When `kreator.makeDataComponent()` is called with a json file, it will create the list of the files with events in certified runs.

The name of the cache file will include the name of the json file.

The advantage of this PR is that we can stop submitting jobs that will process MiniAOD files with no certified data. Note that there are many more MiniAOD files with no certified data than MiniAOD files with certified data.


gpetruc:
merged in the new branch in #580, will close this

PR #528, opened 2015-09-26
CMG local adding Run C&D json, HLT paths and RunD data set

safarzad:
update HLT paths based on RunC and D menu, 25ns json and Run D datasets


mariadalfonso:
@safarzad 
I will close this one:
1) not sure we want to duplicate the file with the trigger list since only a few will change.
2) the runD list wasn't add completely [need to add to the list of dataSamples so that is tested] now I added in #529 together with the inital runD json

PR #540, opened 2015-10-08
Type1 MET corrections in python (and cleanup of JetReCalibrator)

gpetruc:
Allow recomputing the type1 MET from raw MET + jets in the JetAnalyzer + METAnalyzer.
- Reproduces the one computed from AOD within numerical precision of MiniAOD (1-5 MeV in the ttbar events I tested). 
- It does not recompute other MET correction levels, nor MET uncertainties.
- It needs to know which CMSSW version was used to create the input pat::MET object, old (&le; 7_4_11) or new (&ge; 7_4_12)

In the process, I also cleaned up the implementation of JetReCalibrator a bit.


gpetruc:
included in #544

PR #541, opened 2015-10-08
Bugfix: allow running without recalibrateJets with v2 lepton quantities

peruzzim:
Separate re-calculation of L1 and L2L3Res JEC is necessary to calculate v2 lepton-jet quantities with lep-aware JEC. If recalibrateJets==False, this does not happen. In this case, this PR sets ptRatiov2=1, ptRelv2=0.


mangano:
Hi Marco, all, 

I'm running some tests for MT2 on a TTbar MC file using:

jetAna.mcGT     = "Summer15_25nsV2_MC" # jec corrections
jetAna.dataGT   = "Summer15_25nsV2_DATA" # jec corrections
jetAna.recalibrateJets = True # True
jetAna.applyL2L3Residual = False # 'Data'

and I get the error:
AttributeError: 'pat::Jet' object has no attribute 'CorrFactor_L1L2L3Res'

It is triggered by the leptonMVA.py module

Any idea?

peruzzim:
Ciao,

 apologies for this problem, it is caused by the new v2 variables that need the separate corrections to be calculated, but were not adequately protected from cases where the user does not run the full JEC in heppy.

PR #542 should fix the problem temporarily, we will revise and test all possible cases this morning.

 Marco

PR #544, opened 2015-10-09
Cleanup of jet correction issues in last PRs

gpetruc:
Merge #540, #541, #542 together cleanly.

MT2 config runs, waiting for a check from @mariadalfonso that the values are also correct.


gpetruc:
Fixed conflict and updated run_susyMT2 as per instructions from @mariadalfonso

PR #550, opened 2015-10-14
Spring15 ele ID MVA WP + SC eta used for electron EA + misc

peruzzim:
Heppy updates:
- Spring15 non-triggering electron ID MVA working points (very loose + additional ones tuned on top of ID and isolation cuts used in electron triggers) are now used in the electron object
- choice of electron EA value in Spring15 switched from electron eta to SuperCluster eta 
- added latest definition of 1/E-1/p from EGM POG (saved in new variable)

plus: added variables used to emulate electron trigger cuts and multilepton cfg update.


PhysicsTools/Heppy/python/physicsobjects/Electron.py
```diff
@@ -29,9 +29,10 @@ def electronID( self, id, vertex=None, rho=None ):
         elif id == "POG_MVA_ID_Trig":     return self.mvaIDTight()
         elif id == "POG_MVA_ID_NonTrig_full5x5":  return self.mvaIDLoose(full5x5=True)
         elif id == "POG_MVA_ID_Trig_full5x5":     return self.mvaIDTight(full5x5=True)
-        elif id == "POG_MVA_ID_Run2_NonTrig_VLoose":   return self.mvaIDRun2("NonTrigPhys14","VLoose")
-        elif id == "POG_MVA_ID_Run2_NonTrig_Loose":    return self.mvaIDRun2("NonTrigPhys14","Loose")
-        elif id == "POG_MVA_ID_Run2_NonTrig_Tight":    return self.mvaIDRun2("NonTrigPhys14","Tight")
+        elif id == "POG_MVA_ID_Run2_NonTrig_VLoose":   return self.mvaIDRun2("NonTrigSpring15","VLoose")
+        elif id == "POG_MVA_ID_Run2_NonTrig_VLooseIdEmu":   return self.mvaIDRun2("NonTrigSpring15","VLooseIdEmu")
+        elif id == "POG_MVA_ID_Run2_NonTrig_VLooseIdIsoEmu":   return self.mvaIDRun2("NonTrigSpring15","VLooseIdIsoEmu")
+        elif id == "POG_MVA_ID_Run2_NonTrig_Tight":    return self.mvaIDRun2("NonTrigSpring15","Tight")
```
gpetruc: I'd suggest to rename the new ones `POG_MVA_ID_Spring15_...` , otherwise you're changing what people get from their exising CFG without any warning and without any chance for them to preserve the old behaviour.

If you want to avoid confusion, maybe you can rename the old ones to `POG_MVA_ID_Phys14_...`, that will force people to change (they'll get an exception otherwise) but they will be able to preserve the old behaviour by configuring the code correctly

PR #557, opened 2015-10-19
Fixes of some susy CFGs after PR #555 and #550

gpetruc:
- add back triggers that were deleted by mistake
- fix some cfgs that were broken by reames of triggers or MVA WPs


gpetruc:
@mariadalfonso @mdunser @safarzad

PR #569, opened 2015-10-30
Non int weight ids

stiegerb:
Handle non integer LHE Weights


gpetruc:
I see this now has conflicts.
Will try remaking it without all the whitespace changes to see if that is enough to solve the conflicts.

gpetruc:
Replaced by #581 (unless problems are found there), closing

PR #572, opened 2015-11-04
VV: Improvements and Statistics tools

bachtis:


bachtis:
Hi Can someone merge this?

gpetruc:
Hi Mike,

We can merge it, but there's a lot of duplication of FastJet things between
your package and
https://github.com/bachtis/cmg-cmssw/blob/VV_7_4_12/PhysicsTools/Heppy/interface/ReclusterJets.h
and some of the features in PyJetToolbox could also be put in a more
central place since they seem related only to making jets out of candidates
and not on VV resonances.

Ciao,

Giovanni

bachtis:
Well I can move it in the future. The problem is that I needed to change the interface to get the job done and I needed to implement things to get the jet areas right for L1 energy corrections + all substructure stuff. So I suggest to merge it and then propose a place to put the improved fastjet interface +PyJetToolbox and I can try to work on it to move it

PR #577, opened 2015-11-13
Monojet synchronisation and update to Spring15 miniAOD v2 

emanueledimarco:
- update the sequences to run on the miniAODv2 
- update the friend trees with new variables to update the selection to the Jamboree one
- update the selection of the signal region with new synch values
- add the inclusive samples of VV (WW,WZ,ZZ) used in monojet


mariadalfonso:
@emanueledimarco  Can you move the CMGTools/MonoXAnalysis/python/samples/samples_monojet_13TeV_74X_signals.py
to the CMGTools/RootTools/python/samples/

emanueledimarco:
@mariadalfonso done (I thought that the specific signal samples could live in the respective package).

PR #581, opened 2015-11-26
Pr569 cleaned

gpetruc:
As #569 but with conflcts solved (and with no whitespace changes).
@stiegerb can you test it, or ask the tHq people to test it?

BTW, this way we still do multiple times the string &rarr; int conversion, it could be worth testing if things can go faster if we do it only once and make a list of (int, float) for the weights in the LHE analyzer, and then call `TH1D::AddBinContent(Int_t bin, Double_t w)`

### DIFF
--- a/CMGTools/TTHAnalysis/python/analyzers/susyCore_modules_cff.py
+++ b/CMGTools/TTHAnalysis/python/analyzers/susyCore_modules_cff.py
@@ -107,8 +107,6 @@ genAna = cfg.Analyzer(
     # Make also the splitted lists
     makeSplittedGenLists = True,
     allGenTaus = False,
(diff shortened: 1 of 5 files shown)


stiegerb:
@gpetruc Tried and tested and seems to work as expected. Did you change anything except for whitespace?

I agree, it's not very efficient. If that's a bottleneck for the overall speed we should change it. It's also not very smart, since we only need the information once per sample to associate the array position of the weight with the id.

Could we produce a simple dictionary of weight_id -> array position in beginLoop() and either write it directly to a histogram or put it in the component such that it can be used in later analyzers? That way we could also add custom information about the meaning of each weight_id.

PR #594, opened 2016-01-05
Updated JEC with variations + direct EOS stageout + misc

peruzzim:
Summary of changes:
- default JEC switched to 25ns_V6
- added JEC files for FASTSIM
- added copies of jetAnalyzer and metAnalyzer for up/down JEC variations
- added pdgId of matched gen lepton and jet partonFlavour
- added possibility of requiring same-sign in ttH lepton skimmer
- direct stageout to EOS when running heppy_batch with remoteCopy
- updated scripts for tree handling on EOS
- minor fixes

PR #595, opened 2016-01-05
Run on SMS scans + updates to plotting and skimming macros

peruzzim:
Summary of changes:
- adaptations for running on SMS scans, including model parsing, mass point per-event and sum-of-weights storage, and tree splitting by mass point
- added precalculated id/iso emulation for electrons
- skimTrees can now veto events (based on run:lumi:evt txt files) and skim friend trees as well
- updates to mcAnalysis/mcPlots/mcDump: rescaling/regrouping datasets, draw uncertainties on MC, normalization fitting, save dump to txt file
- updated fake rate application functions and multiIso working points
- added functionalities to SUSY datacard preparation script

Follows PR #594.

### DIFF
--- a/CMGTools/TTHAnalysis/python/analyzers/ttHhistoCounterAnalyzer.py
+++ b/CMGTools/TTHAnalysis/python/analyzers/ttHhistoCounterAnalyzer.py
@@ -13,11 +13,14 @@ from PhysicsTools.Heppy.analyzers.core.AutoHandle import AutoHandle
 
 import ROOT
 import os
+import re
 
 class ttHhistoCounterAnalyzer( Analyzer ):
     def __init__(self, cfg_ana, cfg_comp, looperName ):
         super(ttHhistoCounterAnalyzer,self).__init__(cfg_ana,cfg_comp,looperName) 
         self.doLHE = getattr(cfg_ana, 'doLHE', True)
+        self.cfg_ana = cfg_ana
+        self.isInitSMS = False
 
     def declareHandles(self):
         super(ttHhistoCounterAnalyzer, self).declareHandles()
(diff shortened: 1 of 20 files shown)


CMGTools/TTHAnalysis/python/analyzers/ttHhistoCounterAnalyzer.py
```diff
@@ -32,23 +35,74 @@ def beginLoop(self, setup):
             setup.services["outputfile"].file.cd()
             self.inputCounter = ROOT.TH1D("Count","Count",1,0,2)
             if self.cfg_comp.isMC:
+                self.maxSMSmass = getattr(self.cfg_ana, 'SMS_max_mass', 3000)
+                self.inputCounterSMS = ROOT.TH3D("CountSMS","CountSMS",int(self.maxSMSmass+1),-0.5,self.maxSMSmass+0.5,int(self.maxSMSmass+1),-0.5,self.maxSMSmass+0.5,1,0,2)
```
gpetruc: shouldn't the creation of theis TH3 go into `initSMS`? (and the same for the following ones)

CMGTools/TTHAnalysis/python/plotter/mcAnalysis.py
```diff
@@ -195,6 +205,8 @@ def getYields(self,cuts,process=None,nodata=False,makeSummary=False,noEntryLine=
             from multiprocessing import Pool
             pool = Pool(self._options.jobs)
             retlist  = pool.map(_runYields, tasks)
+            pool.close()
+            pool.join()
```
gpetruc, line 209: Why do we need this? from the docs, pool.map already blocks until the tasks are done successfully ( https://docs.python.org/2/library/multiprocessing.html#module-multiprocessing.pool )
is it to solve the hangs that sometimes happen?

PR #601, opened 2016-01-08
Updates from AlphaT

aelwood:
- Updated jet smearing parameters  
- Added options for photon rho area and gen dR matching  
- Added counters to trigger bit filter 
- Added gen photon ntuple type
- Added module to calculate parameters for top pt reweighting


mariadalfonso:
@aelwood can you do the commit in the heppy  branch ?

aelwood:
Hi @mariadalfonso , the PR is made https://github.com/CERN-PH-CMG/cmg-cmssw/pull/602

gpetruc:
Hi @aelwood, 
I see that now the PhotonAnalyzer has two different definition of `isPrompt`, one from the `isPromptFinalState()` of the `GenParticle` and one from https://github.com/aelwood/cmg-cmssw/blob/RA1-CMGTools-from-CMSSW_7_4_12/PhysicsTools/Heppy/python/analyzers/objects/PhotonAnalyzer.py#L160-L162 
Is there a need for both? 
If yes, can they be named differently to avoid confusion?
(adding also @pandolf )

lucien1011:
Hi @gpetruc , I think the isPromptFinalState is not needed and we can remove it.

PhysicsTools/Heppy/python/analyzers/objects/PhotonAnalyzer.py
```diff
@@ -135,7 +152,48 @@ def idWP(gamma,X):
...
         if foundPhoton: self.counters.counter('events').inc('has >=1 gamma at preselection')
         if len(event.selectedPhotons): self.counters.counter('events').inc('has >=1 selected gamma')
-       
+
+    def checkGenPhoton(self, event):
+        def isHardParton(particle):
+            return abs(particle.pdgId()) in [1,2,3,4,5,6,21] and particle.status() == 23
+
+        def isPromptPhoton(photon):
+            return photon.status() == 1 and abs(photon.mother(0).pdgId()) <= 22 and abs(photon.mother(0).pdgId()) not in [11,13,15]
```
gpetruc, line 161: Here, FSR photons from prompt leptons will not be flagged as prompt, while the `GenParticle` status flags like  `isPromptFinalState` would call them prompt (and they would be isolated, so calling them prompt is probably right).
Maybe can you call it `isPromptPhotonNoFSR` or something similar?
Or, what are you trying to achieve?
I think it would be good to have a more solid definition of what we call mc truth photons in heppy, and if for whatever reason we need a more ad-hoc definition because it's used in susy then it can be put somewhere else (or labelled somehow)
Tagging @bendavid which just told me that not all FSR photons will have a lepton as a mother if they're added by Madgraph and not Pythia

bendavid, line 161: (Technically ISR and FSR photons can even interfere, so checking the mother of the photon in this way is not well defined). 

If you really must distinguish between "photons radiated from final state leptons" and other prompt photons, the more robust way would actually be to check delta R to the nearest status 1 lepton.  What is the motivation to call these photons non-prompt?

PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
```diff
@@ -387,15 +387,16 @@ def smearJets(self, event, jets):
...
+               #https://twiki.cern.ch/twiki/bin/view/CMS/JetResolution#Smearing_procedures
                
-               factor = 1.079 + self.shiftJER*0.026
-               if   aeta > 3.2: factor = 1.056 + self.shiftJER * 0.191
-               elif aeta > 2.8: factor = 1.395 + self.shiftJER * 0.063
-               elif aeta > 2.3: factor = 1.254 + self.shiftJER * 0.062
-               elif aeta > 1.7: factor = 1.208 + self.shiftJER * 0.046
-               elif aeta > 1.1: factor = 1.121 + self.shiftJER * 0.029
-               elif aeta > 0.5: factor = 1.099 + self.shiftJER * 0.028
+               factor = 1.061 + self.shiftJER*0.023
```
mariadalfonso: @aelwood.
there is a conflict on this with the HEAD after the #600 me, move the 13TeV parameters in the function shiftJERfactor
https://github.com/CERN-PH-CMG/cmg-cmssw/pull/600/files#diff-b4f7bda6e988f3a66e98f9d2f807d14cR51

aelwood:
@mariadalfonso the merge conflict should be fixed now

gpetruc:
Closing, I think that wherever we go probably this PR is by now obsolete.

PR #608, opened 2016-01-19
E/gamma scale calibration & smearing (74X)

gpetruc:
https://twiki.cern.ch/twiki/bin/view/CMS/EGMSmearer


gpetruc:
Note for the future: the 76X PR will have to be based on the 76X e/gamma branch, not the 74X one.
So, we should just start from the e/gamma branch and then cherry pick  100d1db  and 79813b0 (updating the default numbers)

gpetruc:
@emanueledimarco is testing it

emanueledimarco:
@gpetruc I tested the updated PR and it is consistent with my old smearer of the supercluster only.

PR #611, opened 2016-01-20
CMGTools: move trigger bit variables from float to int

brucala:
related to heppy PR #610


gpetruc:
This and #610 had the titles swapped, I have corrected them now.

PR #614, opened 2016-01-21
VV plotting update

mtosi:
add ratio plot and fix CMS text

### DIFF
--- a/CMGTools/VVResonances/python/plotting/StackPlotter.py
+++ b/CMGTools/VVResonances/python/plotting/StackPlotter.py
@@ -64,7 +64,7 @@ def CMS_lumi(lumi,pad):
         print lumi
 
     lumiText = "#scale[0.85]{" + str(lumi) + " fb^{-1} (13 TeV) }"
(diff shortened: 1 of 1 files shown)


gpetruc:
@bachtis do you want to review the PRs on the VV subsystem?

bachtis:
Hi @gpetruc . In fact , now that we are more than one in the VV subsystem I would like to merge . But we talk with Mia so no need to do +1s . Let me know if I can accept merges that are independent of other packages

mtosi:
ciao
I've just committed the update
we have 2 distinct functions, now
drawStack and drawStackAndRatio

the latter is the new one w/ both stack and ratio
the former is almost the same as the original one

gpetruc:
Ok. 

Mike, now you have merge rights. 

If it's something that affects **only** VV, you can go ahead and merge.
If it's something that affects both VV and something else (e.g. if there are Heppy commits), I would also check the things before, so wait for a message from me before merging.

bachtis:
ok! thanks

PR #617, opened 2016-02-02
MonoX control samples updates

emanueledimarco:
- several changes in the selection for the 4 lepton control regions
- changes to use the pt-dependent control regions
- add a script to check the sanity of friend trees


CMGTools/MonoXAnalysis/python/tools/eventVars_monojet.py
```diff
@@ -3,8 +3,8 @@
 from CMGTools.MonoXAnalysis.tools.BTagWeightCalculator import BTagWeightCalculator
```
mariadalfonso: @emanueledimarco 
Why the BTagWeightCalculator is inside the MonoXAnalysis ? IS there something specific, there are other duplicate in other packages.

emanueledimarco: Yes, this is a duplicate of a module that @gpetruc gave me, and it should be an updated version from H->bb of this file : 
 https://github.com/CERN-PH-CMG/cmg-cmssw/blob/ea63f1bcfbf798253aa97fd8814ce7871b83d801/PhysicsTools/Heppy/python/physicsutils/BTagWeightCalculator.py

if experts of that can comment, we could update that file in heppy and I will just import from there.

emanueledimarco:
@mariadalfonso 
With @gpetruc we checked that the BTagWeightCalculator seems to be newer than the version in heppy. Also in the heppy_76X branch the version was the same. 
Then I updated the PR updating the file in Heppy, removing the one in the CMGTools/MonoXanalysis.

PR #626, opened 2016-03-29
First commit for the MT2 in 76X

mmasciov:


PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
```diff
@@ -238,10 +239,12 @@ def process(self, event):
...
                 photons = [ g for g in event.selectedPhotons ] 
 
-        self.gamma_cleanJetsAll = cleanNearestJetOnly(self.cleanJetsAll, photons, self.jetGammaDR)
+#        self.gamma_cleanJetsAll = cleanNearestJetOnly(self.cleanJetsAll, photons, self.jetGammaDR)
+        self.gamma_cleanJetsAll = cleanNearestJetOnly(jetsEtaCut, photons+leptons, self.jetGammaDR)
         self.gamma_cleanJets    = [j for j in self.gamma_cleanJetsAll if abs(j.eta()) <  self.cfg_ana.jetEtaCentral ]
         self.gamma_cleanJetsFwd = [j for j in self.gamma_cleanJetsAll if abs(j.eta()) >= self.cfg_ana.jetEtaCentral ]
-        self.gamma_noIdCleanJetsAll = cleanNearestJetOnly(self.noIdCleanJetsAll, photons, self.jetGammaDR)
+#        self.gamma_noIdCleanJetsAll = cleanNearestJetOnly(self.noIdCleanJetsAll, photons, self.jetGammaDR)
+        self.gamma_noIdCleanJetsAll = cleanNearestJetOnly(self.jetsAllNoID, photons+leptons, self.jetGammaDR)
```
gpetruc: This way you use `jetGammaDR` also for the leptons, which doesn't seem right to me (even if in practice they're likely to be both `0.4`)

mmasciov: Yes, this was actually done on purpose, since we want to clean jets from leptons and photons together (as a single collection equal to the sum of the two), with the same dR (=0.4). 
However, I agree this might be ambiguous.
I have just made a new commit, where I'm still doing the same but defining a different dR variable, called jetGammaLepDR. This way, the purpose of the cleaning should be non-ambiguous. 
Also, the jet/lepton cross-cleaning is unchanged, and it's done using jetLepDR, as usual. So, this change should not affect anyone else (I think MT2 is the only one using the gamma_jet collection?).

gpetruc: I did use the jet cleaned from the photons for a while in H->ZZ for example, and it may be used e.g. by the mono-photon people (the pool of heppy users is much larger than just the 3-4 susy analyses).

The new commit is basically a rename of the `jetGammaDR` to `jetGammaLepDR` since nothing else in the code afterwards uses `jetGammaDR`, so it doesn't really change anything besides leaving an extra configurable parameter that does nothing, so I'd drop it.

Two possible solutions:
- add a switch `cleanFromLepAndGammaSimultaneously`: if true, use the logic of this PR and raise an exception at configuration time if the two deltaR's are different; if false, stay with the old (first clean from the leptons with the lepton DR, then clean from the jets)
- make an extended cleanNearestJetOnly that it can use different deltaR's for the two classes of objects. 

The first option seems preferable to me

mmasciov: I implemented an option cleanFromLepAndGammaSimultaneously: if true, clean jets from lepton and photons simultaneously; if false, the old logic is kept.
If true, and jetGammaLepDR is not set in the cfg file: if jetGammaDR==jetLepDR take the same value, otherwise take 0.4 (default) and just print a warning. Let me know if this looks good to you. Thanks

PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
```diff
@@ -238,10 +240,26 @@ def process(self, event):
...
+        self.gamma_cleanJetaAll = []
+        self.gamma_noIdCleanJetsAll = []
+
+        if getattr(self.cfg_ana, 'cleanFromLepAndGammaSimultaneously', False):
+            if hasattr(self.cfg_ana, 'jetGammaLepDR'):
+                self.jetGammaLepDR =  self.cfg_ana.jetGammaLepDR 
+            elif (self.jetGammaDR == self.jetLepDR):
+                self.jetGammaLepDR = jetGammaDR
+            else:
+                print "WARNING: dR for simultaneous cleaning of jets from leptons and photons is not defined, and dR(gamma, jet)!=dR(lep, jet). Using default (0.4)"
```
gpetruc: Last small requests before merging this PR: 
- add the `cleanFromLepAndGammaSimultaneously` to the default configuration at the bottom
- move this part of the check, that happens at configuration time, to the constructor of the analyzer and change the `print message` with `raise RuntimeError, message`

mmasciov: This is now done and committed. Thanks

PR #629, opened 2016-04-11
Protecting Jet for case one wants to run Jet analyzer on AK8

bachtis:
In AK8 jets there is no userfloat for puJetID so the JetAna crashes
if the userfloat doesnt exist I modified the class to give -9


bachtis:
What is going to happen with that?

PR #635, opened 2016-05-03
Fix to energyFraction methods in Jet class

mangano:
The energyFractions methods are redefined within the python Jet class to avoid non-sense values from the C++ methods when extra energy corrections are applied

### DIFF
--- a/PhysicsTools/Heppy/python/physicsobjects/Jet.py
+++ b/PhysicsTools/Heppy/python/physicsobjects/Jet.py
@@ -49,6 +49,48 @@ class Jet(PhysicsObject):
         self._leadingTrack = None
         self._leadingTrackSearched = False
 
+    def rawEnergy(self):
+        return (self.p4()*self.rawFactor()).energy()
+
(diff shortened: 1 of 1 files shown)


PhysicsTools/Heppy/python/physicsobjects/Jet.py
```diff
@@ -49,6 +49,48 @@ def _physObjInit(self):
         self._leadingTrack = None
         self._leadingTrackSearched = False
 
+    def rawEnergy(self):
+        return (self.p4()*self.rawFactor()).energy()
```
gpetruc: Why not `self.energy() * self.rawFactor()` avoiding to create a 4-vector object for nothing?

mangano: I was lazy and copied that line from the already existing jetID() method. Now both the new rawEnergy method and the jetID one are fixed.

gpetruc:
Ok, merged now. Can you also rebase to `heppy_80X` ?

PR #637, opened 2016-05-04
Fix memory leak in jet analyzer

mmarionncern:
The copy-by-value option in the jet analyzer was memory leaking due to a pyROOT feature.

The copy is now embedded into a C++ function. Few tests do not show any leak anymore

### DIFF
--- a/PhysicsTools/Heppy/interface/JetUtils.h
+++ b/PhysicsTools/Heppy/interface/JetUtils.h
@@ -1,0 +1,18 @@
+#ifndef PhysicsTools_Heppy_JetUtils_h
+#define PhysicsTools_Heppy_JetUtils_h
+
+#include "DataFormats/Common/interface/Ptr.h"
+#include "DataFormats/PatCandidates/interface/Jet.h"
+
+namespace heppy{
+
+  struct JetUtils{
+
+    static const pat::Jet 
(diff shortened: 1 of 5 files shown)


PhysicsTools/Heppy/interface/JetUtils.h
```diff
@@ -0,0 +1,18 @@
...
+
+#include "DataFormats/Common/interface/Ptr.h"
+#include "DataFormats/PatCandidates/interface/Jet.h"
+
+namespace heppy{
+
+  struct JetUtils{
+
+    static const pat::Jet 
+    copyJet(const edm::ProductID& id, const pat::Jet& ijet, unsigned long key);
```
gpetruc: I'd suggest to have this just be `copyJet(const pat::Jet &jet)` and hide also the temporary `edm::Ptr` inside the function.
It should be both cleaner to read and faster.

mmarionncern: By hiding the edm::Ptr, you mean in a new pat::Jet constructor? (not sure about what you have in mind, as it is already in the C++ function)

gpetruc:
```
static const pat::Jet copyJet(const pat::Jet& ijet) {
     edm::Ptr<pat::Jet> ptr(edm::ProductID(), &ijet, 0);
     return pat::Jet(ptr);
}
```

PR #639, opened 2016-05-23
Update of photon working points

aelwood:
It looks like there may have been a decimal place error for the photon working point, the twiki has numbers such as 0.014, the code has 0.0014 . ( https://twiki.cern.ch/twiki/bin/viewauth/CMS/CutBasedPhotonIdentificationRun2#SPRING15_selections_25_ns ) .

Additionally added the photon isolation to the spring15 WP.

Thanks to @lucien1011 for spotting.


gpetruc:
Should there be also a backport to 76X?

aelwood:
@gpetruc yes you're right, I've made it here https://github.com/CERN-PH-CMG/cmg-cmssw/pull/643

PR #641, opened 2016-05-23
add options in the alphat calculation

TaiSakuma:
A PR from the RA1 private code.

This PR adds options in the alphat calculation:
- an option to attache pseudo-jet flags to jets
- an option to use pT instead of ET


gpetruc:
Will we need a backport to 76X as well, if 2015 data has to be re-analyzed?

PR #642, opened 2016-05-26
Workaround for memory leak in GenAnalyzer

gpetruc:
Apparently the method `motherRef` of the `reco::GenParticle` class leaks memory within ROOT/Cling type resolution. This PR fixes it by getting the key of the reference in C++ instead of python.
Added also the option `--memcheck` to heppy to turn on memory checking from command line.

Similar fixes to some other CMG modules will also be needed.

@arizzi @mmarionncern @peruzzim @mangano

### DIFF
--- a/PhysicsTools/HeppyCore/python/framework/heppy_loop.py
+++ b/PhysicsTools/HeppyCore/python/framework/heppy_loop.py
@@ -45,12 +45,14 @@ def runLoop( comp, outDir, config, options):
     fullName = '/'.join( [outDir, comp.name ] )
     # import pdb; pdb.set_trace()
     config.components = [comp]
+    memcheck = 2 if getattr(options,'memCheck',False) else 0
     loop = Looper( fullName,
                    config,
                    options.nevents, 0,
                    nPrint = options.nprint,
                    timeReport = options.timeReport,
-                   quiet = options.quiet)
+                   quiet = options.quiet,
+                   memCheckFromEvent = memcheck)
     # print loop
     if options.iEvent is None:
         loop.loop()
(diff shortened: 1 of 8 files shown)


mangano:
Hi Giovanni, 

I tested these 3 commits in my setup and the memory leak seems to be gone to first approximation: RES memory from top stable at ~800 MB instead of jumping to 1.5G after just 5 mins of running. 

However, I get these messages here:
Mem Jump detected in analyzer treeProducerSusyFullHad at event 2333. RSS(before,after,difference) 828416 828420 4 
Mem Jump detected in analyzer treeProducerSusyFullHad at event 2336. RSS(before,after,difference) 828420 828424 4 
Mem Jump detected in analyzer treeProducerSusyFullHad at event 2341. RSS(before,after,difference) 828424 828428 4 

Is this due to the similar (may be smaller) memory leaks that you referred to with "Similar fixes to some other CMG modules will also be needed." ? 

Thanks,
Boris

gpetruc:
The treeProducer may have a legitimate growing memory footprint e.g. because of buffers that are not flushed until the end of the job or so, so I wouldn't worry about it unless it explodes.
The CMG fixes I was referring to are being collected in https://github.com/CERN-PH-CMG/cmgtools-lite/pull/44 (there may be others that I haven't found yet)

PhysicsTools/HeppyCore/python/framework/heppy_loop.py
```diff
@@ -45,12 +45,14 @@ def runLoop( comp, outDir, config, options):
     fullName = '/'.join( [outDir, comp.name ] )
     # import pdb; pdb.set_trace()
     config.components = [comp]
+    memcheck = 2 if getattr(options,'memCheck',False) else 0
```
mangano, line 48: Hi, 
If I understand correctly the logic of the Looper class, in order to have the mem checker being inactive by default this line should be like this:

memcheck = 2 if getattr(options,'memCheck',False) else -1

no ?

gpetruc, line 48: Yes... will fix ASAP

PR #646, opened 2016-05-27
Heppy updates for multilepton

peruzzim:
- variables for soft lepton ID
- new version of lepton generator match
- more tau ID working points


PhysicsTools/Heppy/python/analyzers/objects/LeptonAnalyzer.py
```diff
@@ -437,8 +459,12 @@ def attachMiniIsolation(self, mu):
         what = "mu" if (abs(mu.pdgId()) == 13) else ("eleB" if mu.isEB() else "eleE")
         if what == "mu":
             mu.miniAbsIsoCharged = self.IsolationComputer.chargedAbsIso(mu.physObj, mu.miniIsoR, {"mu":0.0001,"eleB":0,"eleE":0.015}[what], 0.0);
+            mu.miniAbsIsoChargedFix03 = self.IsolationComputer.chargedAbsIso(mu.physObj, 0.3, {"mu":0.0001,"eleB":0,"eleE":0.015}[what], 0.0);
+            mu.miniAbsIsoChargedFix04 = self.IsolationComputer.chargedAbsIso(mu.physObj, 0.4, {"mu":0.0001,"eleB":0,"eleE":0.015}[what], 0.0);
         else:
             mu.miniAbsIsoCharged = self.IsolationComputer.chargedAbsIso(mu.physObj, mu.miniIsoR, {"mu":0.0001,"eleB":0,"eleE":0.015}[what], 0.0,self.IsolationComputer.selfVetoNone);
+            mu.miniAbsIsoChargedFix03 = self.IsolationComputer.chargedAbsIso(mu.physObj, 0.3, {"mu":0.0001,"eleB":0,"eleE":0.015}[what], 0.0,self.IsolationComputer.selfVetoNone);
+            mu.miniAbsIsoChargedFix04 = self.IsolationComputer.chargedAbsIso(mu.physObj, 0.4, {"mu":0.0001,"eleB":0,"eleE":0.015}[what], 0.0,self.IsolationComputer.selfVetoNone);
```
gpetruc: Why do we have these, and why are they not optional? plus, they shouldn't be called miniWhateverIso since they're not mini, they're just the standard isolation cones recomputed.

PhysicsTools/Heppy/python/analyzers/objects/LeptonAnalyzer.py
```diff
@@ -187,9 +194,24 @@ def makeLeptons(self, event):
             if self.miniIsolationVetoLeptons == "inclusive":
                 for lep in event.inclusiveLeptons: 
                     self.IsolationComputer.addVetos(lep.physObj)
+            event.LeptonTrackMuPairs = []
             for lep in event.inclusiveLeptons:
                 self.attachMiniIsolation(lep)
-        
+                if self.doDirectionalIsolation:
+                    for cs in xrange(2,7): self.attachDirectionalIsolation(lep,0.1*cs)
```
gpetruc: Maybe it would be better to have `doDirectionalIsolation` be equal to the list of cone sizes to compute, so that we don't recompute 5 isolation cones every time?

PhysicsTools/Heppy/python/analyzers/objects/LeptonAnalyzer.py
```diff
@@ -596,6 +667,67 @@ def distance(rec,gen):
...
+
+    def isFromGamma(self,particle,gid=22, done={}):
+        for i in xrange( particle.numberOfMothers() ): 
+            mom  = particle.mother(i)
+            momid = abs(mom.pdgId())
+            if momid == gid: 
+                return True
+        return False
+
+    def SUSYMatchLeptons(self, event):
```
gpetruc: I think this should be relocated to a separate analyzer.

PhysicsTools/Heppy/python/analyzers/objects/LeptonAnalyzer.py
```diff
@@ -187,9 +194,24 @@ def makeLeptons(self, event):
...
+                    for pairedCand in self.IsolationComputer.findPairIsoTrack(lep.physObj, 999, 0.0, 4.0, 1.5, 5.0):
+                        _lep_p4 = lep.p4()
+                        _pc_p4 = pairedCand.p4()
+                        lep_p4 = TLorentzVector(_lep_p4.Px(),_lep_p4.Py(),_lep_p4.Pz(),_lep_p4.E())
+                        pc_p4 = TLorentzVector()
+                        pc_p4.SetPtEtaPhiM(_pc_p4.Pt(),_pc_p4.Eta(),_pc_p4.Phi(),0.105)
+                        self.pairedIsoChargedCandidatesMuHyp.append((pairedCand,(lep_p4+pc_p4).M()))
+                    self.pairedIsoChargedCandidatesMuHyp.sort(key = lambda x : x[1], reverse = True)
+                    event.LeptonTrackMuPairs.extend([(lep,x[0],x[1]) for x in self.pairedIsoChargedCandidatesMuHyp])
+                    event.LeptonTrackMuPairs.sort(key = lambda x : x[2], reverse = True)
```
gpetruc: This feature also doesn't seem to me to be so general, it would better be kept in another analyzer

PhysicsTools/Heppy/src/IsolationComputer.cc
```diff
@@ -14,6 +14,17 @@ namespace {
...
+        bool operator()(const pat::PackedCandidate *c1, const pat::PackedCandidate *c2) const {
+            return c1->pt() < c2->pt();
+        }
+        bool operator()(float c1pt, const pat::PackedCandidate *c2) const {
+            return c1pt < c2->pt();
+        }
+        bool operator()(const pat::PackedCandidate *c1, float c2pt) const {
+            return c1->pt() < c2pt;
+        }
+    };
```
gpetruc, line 21: Only the first of the three methods is needed

PhysicsTools/Heppy/src/IsolationComputer.cc
```diff
@@ -129,6 +158,43 @@ float heppy::IsolationComputer::isoSumRaw(const std::vector<const pat::PackedCan
     return isosum;
 }
 
+reco::Candidate::LorentzVector heppy::IsolationComputer::isoP4Raw(const std::vector<const pat::PackedCandidate *> & cands, const reco::Candidate &cand, float dR, float innerR, float threshold, SelfVetoPolicy selfVeto, int pdgId) const
```
gpetruc: If these are copies of the non-P4 methods except that they add up cand->p4() rather than cand->pt(), then it would be better to make a template and call it (just like the `algoWithin` in DataFormats/RecoCandidate/interface/IsoDeposit.h )

PR #649, opened 2016-06-02
Updated PhotonID bits and updated Jet PU ID (76X WP tuning)

emanueledimarco:
- Updated the bits of the photon ID (including the medium one) taking the values from the Spring15 25ns cut-based selections computed in heppy instead using the values stored in the photon object which is obsolete
- updated the PU jet ID to a new WP tuning done on top of the MVA training in 76X. This will be redone in 80X, but this should be better than the cuts that were tuned on the 53X 
  - the 53X version is still there, and can be used

### DIFF
--- a/PhysicsTools/Heppy/python/analyzers/objects/PhotonAnalyzer.py
+++ b/PhysicsTools/Heppy/python/analyzers/objects/PhotonAnalyzer.py
@@ -112,8 +112,10 @@ class PhotonAnalyzer( Analyzer ):
                     id=3
                 return id
 
-            gamma.idCutBased = idWP(gamma, "POG_SPRING15_25ns_%s")
-
+            # bits to store in the trees
+            gamma.idWPs = idWP(gamma, "POG_SPRING15_25ns_%s")
+            # bit to be used in selection
+            gamma.idCutBased = 0
 
             keepThisPhoton = True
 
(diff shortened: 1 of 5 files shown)


gpetruc:
Looks ok to me.

To SUSY Hadronic people: any of you need `gamma.idCutBased` to have a backwards-compatible switch, or it's ok to just update it?  @mmasciov @pandolf @TaiSakuma @aelwood @mangano

lucien1011:
@emanueledimarco Can you explain a bit more about the change on the photon? I think gamma.idCutBased will always have the value 1 because a non-empty dictionary of the working point defintion will be returned by CutBasedIDWP.

mmasciov:
@gpetruc @emanueledimarco @pandolf @mangano 
It looks to me that if one is using a 'NoIso' photon ID, then gamma.idCutBased is overwritten: L132 of PhotonAnalyzer.py overwrites L115, which is the line that was changed, doesn't it?

We are using a 'NoIso' ID, therefore if I'm interpreting the code correctly this change should be transparent for us.

pandolf:
just had a chat with @emanueledimarco and i can confirm that this PR is fine from the MT2 point of view.

emanueledimarco:
yes, but I discovered that what I did was not what I needed.
So going to update it soon and re-check with @pandolf

emanueledimarco:
- @gpetruc @pandolf done. The gamma.idCutBased set later and used in the photon selection remains untouched, while in the trees (autophoby.py) the value of idWPs is saved (which contains the loose/medium/tight ID values).
- since I touched the autophobj.py, I also added the supercluster eta to the photon variables

emanueledimarco:
I just updated the PR with:
- an update to the effective areas for photon isolation for Spring15 25ns. The old ones are selected if the gammaID parameter of the photonAnalyzer is "POG_PHYS14_25ns", otherwise the new ones are taken by default
- a bug fix in the recently changed bits of the photon ID stored in the trees

I tested the PR on a synch exercise and for me it can be merged.

gpetruc:
Is @pandolf still OK with this PR?

pandolf:
yep.

PR #657, opened 2016-06-15
Heppy 80X upstream changes

gpetruc:
This brings in the upstream changes to Heppy

The only non-trivial conflict was in the Jet `puJetId` method:
- CMG change: 13a16750b51a24be785db6a1942b9132f14967d3  from #649  (76X WPs)
- VHbb change: 28ecf7a88753445b59c9bdde31719690f56089c7 (allow an external pileup jetId `ValueMap`)
  - Resolved result: https://github.com/gpetruc/cmg-cmssw/blob/cbernet_heppy_8011/PhysicsTools/Heppy/python/physicsobjects/Jet.py#L143-L160

There is a new required parameters in **JetAnalyzer**:  `propagateJetSmearToMET`, a boolean whose value that matters only if you do jet smearing

Susy multilepton validation shows only a few small differences in the jet `mcMatchId` variable, all the rest (including all data) seems to be unaffected.

Tau information has been added and renamed, see 5f9d06ed39cad103682cf510938e1d61cc3db873

Everybody is encouraged to test this PR and report here.


PhysicsTools/Heppy/python/analyzers/objects/TauAnalyzer.py
```diff
@@ -81,14 +81,16 @@ def id6(tau,X):
...
 
-            tau.idMVA = id6(tau, "by%sIsolationMVArun2v1DBoldDMwLT")
-            tau.idMVANewDM = id6(tau, "by%sIsolationMVArun2v1DBnewDMwLT")
+            tau.idMVArun2 = id6(tau, "by%sIsolationMVArun2v1DBoldDMwLT")
+            tau.idMVArun2dR03 = id6(tau, "by%sIsolationMVArun2v1DBdR03oldDMwLT")
+            tau.idMVArun2NewDM = id6(tau, "by%sIsolationMVArun2v1DBnewDMwLT")
             tau.idCI3hit = id3(tau, "by%sCombinedIsolationDeltaBetaCorr3Hits")
+#            tau.idCI3hitdR03 = id3(tau, "by%sCombinedIsolationDeltaBetaCorr3HitsdR03")
             tau.idMVAOldDMRun2 = id3(tau, "by%sIsolationMVArun2v1DBoldDMwLT")
             tau.idMVAOldDMRun2dR03 = id3(tau, "by%sIsolationMVArun2v1DBdR03oldDMwLT")
```
steggema, line 90: This is the same discriminator as in line 85, but here not all 6 working points are saved, so I'd suggest to remove it here.

PhysicsTools/Heppy/python/analyzers/objects/TauAnalyzer.py
```diff
@@ -81,14 +81,16 @@ def id6(tau,X):
...
                 return id5(tau, X) + tau.tauID(X%"VVTight")
 
-            tau.idMVA = id6(tau, "by%sIsolationMVArun2v1DBoldDMwLT")
-            tau.idMVANewDM = id6(tau, "by%sIsolationMVArun2v1DBnewDMwLT")
+            tau.idMVArun2 = id6(tau, "by%sIsolationMVArun2v1DBoldDMwLT")
+            tau.idMVArun2dR03 = id6(tau, "by%sIsolationMVArun2v1DBdR03oldDMwLT")
+            tau.idMVArun2NewDM = id6(tau, "by%sIsolationMVArun2v1DBnewDMwLT")
             tau.idCI3hit = id3(tau, "by%sCombinedIsolationDeltaBetaCorr3Hits")
+#            tau.idCI3hitdR03 = id3(tau, "by%sCombinedIsolationDeltaBetaCorr3HitsdR03")
             tau.idMVAOldDMRun2 = id3(tau, "by%sIsolationMVArun2v1DBoldDMwLT")
```
steggema, line 89: Same comment as next line (same discriminator as in line 84)

PhysicsTools/Heppy/python/analyzers/objects/TauAnalyzer.py
```diff
@@ -81,14 +81,16 @@ def id6(tau,X):
                     loose, medium, tight, very tight, very very tight)"""
                 return id5(tau, X) + tau.tauID(X%"VVTight")
 
-            tau.idMVA = id6(tau, "by%sIsolationMVArun2v1DBoldDMwLT")
-            tau.idMVANewDM = id6(tau, "by%sIsolationMVArun2v1DBnewDMwLT")
+            tau.idMVArun2 = id6(tau, "by%sIsolationMVArun2v1DBoldDMwLT")
```
steggema, line 84: Overall on the naming: 
At the moment, the MVA is e.g. called "MVArun2v1DBoldDMwLT". The two things that vary with the current discriminators are "with and without lifetime variables", and "oldDM vs newDM". With an updated training, I'd expect the version to increase from v1 to v2. Rather than adding "run2" to the names, we may want to let the names reflect what may actually change (the previous idea was to assume that "wLT" and "oldDM" are the defaults, and deviations from this would be reflected in the name). I'll leave it to you to decide - if you think having "run2" in the name improves the clarity, I'm also fine with it.

PhysicsTools/Heppy/python/analyzers/objects/TauAnalyzer.py
```diff
@@ -81,14 +81,16 @@ def id6(tau,X):
...
+            tau.idMVArun2NewDM = id6(tau, "by%sIsolationMVArun2v1DBnewDMwLT")
             tau.idCI3hit = id3(tau, "by%sCombinedIsolationDeltaBetaCorr3Hits")
+#            tau.idCI3hitdR03 = id3(tau, "by%sCombinedIsolationDeltaBetaCorr3HitsdR03")
             tau.idMVAOldDMRun2 = id3(tau, "by%sIsolationMVArun2v1DBoldDMwLT")
             tau.idMVAOldDMRun2dR03 = id3(tau, "by%sIsolationMVArun2v1DBdR03oldDMwLT")
             tau.idAntiMu = tau.tauID("againstMuonLoose3") + tau.tauID("againstMuonTight3")
-            tau.idAntiE = id5(tau, "againstElectron%sMVA6")
-            #print "Tau pt %5.1f: idMVA2 %d, idCI3hit %d, %s, %s" % (tau.pt(), tau.idMVA2, tau.idCI3hit, tau.tauID(self.cfg_ana.tauID), tau.tauID(self.cfg_ana.tauLooseID))
+#            tau.idAntiE = id5(tau, "againstElectron%sMVA5")
+            tau.idAntiErun2 = id5(tau, "againstElectron%sMVA6")
```
steggema, line 93: Similar comment as above: If you want to add anything to the name, I'd rather add the version number ("6" in this case) than the generic "run2".

gpetruc:
Hi Jan,

I agree with your comments on the Tau side. Can you settle this with @veelken , who made those changes in the Heppy? 

Giovanni

steggema:
Hi Giovanni,

Yes I will.

Jan

veelken:
Hi,

I agree with Jan that lines 89 and 90 of https://github.com/gpetruc/cmg-cmssw/blob/d9cc7a27f08447237e94c46c4dfc3e76d2a7e881/PhysicsTools/Heppy/python/analyzers/objects/TauAnalyzer.py can be removed. Concerning the name, I would personally prefer a more descriptive name than the person who trained the MVA ("Arun"). I am fine with using MVAv6 for the discriminator against electrons.

Cheers,

Christian

gpetruc:
will close, and re-open a new one based on 8_0_19

PR #659, opened 2016-06-27
adding ctagger disc, Heppy

safarzad:
adding two c-tagger discriminators


gpetruc:
Is it becoming widely used? otherwise, I'd suggest to add it only to jetTypeSusy or whatever type you use in the analysis, and to jetTypeExtra

PR #662, opened 2016-07-06
Add 80X pileup jet ID and fix 76X pileup jet ID

steggema:


steggema:
Seeing only now that there are also changes to the pileup jet ID in #657 but I think these are orthogonal

PR #665, opened 2016-07-15
use the new Kalman filter muon scale corrections

emanueledimarco:
Update the interface to the updated Kalman filter muon scale corrections from Mike.
See the twiki:

```
https://twiki.cern.ch/twiki/bin/view/CMS/MuonScaleResolKalman
```

for instructions.


gpetruc:
@bachtis do we have a KaMuCa tag with the correction of the ptError yet?

Giovanni

bachtis:
It should be ready by the end of the day

PR #667, opened 2016-07-27
Added N_ISR jets to event for SUSY signal model systematics

aelwood:
The N_ISR variable for ISR signal model systematics was introduced in a SUSY meeting in the following talk: https://indico.cern.ch/event/557678/contributions/2247944/attachments/1311994/1963568/16-07-19_ana_manuelf_isr.pdf

A sample of code was provided by @manuelfs in: https://github.com/manuelfs/babymaker/blob/0136340602ee28caab14e3f6b064d1db81544a0a/bmaker/plugins/bmaker_full.cc#L1268-L1295

This has been converted to python and implemented in the Jet analyzer here.

### DIFF
--- a/PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
+++ b/PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
@@ -443,13 +443,9 @@ class JetAnalyzer( Analyzer ):
                 momid = abs(mc.mother().pdgId())
                 if not (momid==6 or momid==23 or momid==24 or momid==25 or momid>1e6): continue
                     #check against daughter in case of hard initial splitting
-                    #for (size_t idau(0); idau < mc.numberOfDaughters(); idau++) {
                 for idau in range(mc.numberOfDaughters()):
-                    #float dR = deltaR(clean_jets[ijet], mc.daughter(idau)->p4());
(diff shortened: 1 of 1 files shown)


PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
```diff
@@ -424,6 +425,31 @@ def isFlavour(x,f):
...
+        for jet in self.cleanJetsAll:
+
+            if jet.pt()<30.0: continue
+            if abs(jet.eta())>2.4: continue
+            matched = False
+            for mc in event.genParticles:
+                if matched: break
+                if (mc.status()!=23 or abs(mc.pdgId())>5): continue
+                momid = abs(mc.mother().pdgId())
+                if not (momid==6 or momid==23 or momid==24 or momid==25 or momid>1e6): continue
```
steggema, line 443: Not sure whether you know but this will e.g. not catch jets from DY events away from the Z peak where no Z boson appears in the event record. I don't know whether this is a practical issue for your analysis, but I'd suggest to use something more generic if you want to put this into common code.

aelwood, line 443: @steggema , thanks for pointing this out. This implementation has only been designed to work for ttbar and SUSY signal models. I will therefore move it to a different, less generic, analyzer

aelwood:
Added to the cmgtools repository instead in a less generic place, https://github.com/CERN-PH-CMG/cmgtools-lite/pull/104

PR #668, opened 2016-09-05
Update to CMSSW 8_0_19

gpetruc:
This PR updates the base CMSSW branch of heppy to CMSSW_8_0_19
This PR does not change Heppy nor EgammaAnalysis, so it should be transparent, but please give it a test.

The next step will be the PR to update to Colin's Heppy in 8_0_19, which you can preview via
https://github.com/gpetruc/cmg-cmssw/compare/heppy_80X_to_8019...cbernet_heppy_8_0_19.


clelange:
I checked 200 events for a VVResonances signal sample and they are identical --> looks good to me.

PR #669, opened 2016-09-07
Upstream heppy updates (8_0_19 version)

gpetruc:
As #657 but now with the source branch updated to 8_0_19.
I put it here so that people can have a look and do some testing - note that it doesn't even merge cleanly at the moment.


mariadalfonso:
@gpetruc What we do with this PR ?

cbernet:
Hi Maria and all,

It would be nice to finalize it.

I'm in the process of syncing heppy with the FCC version. There are quite a few changes coming in PhysicsTools/HeppyCore. It would be easier for you to do your PR to me on my heppy_8_0_19 (and thus to first deal with with this merge). When this is done I can do my merge with the FCC stuff, and then deal with the porting to the most recent CMSSW version. 

Cheers,

Colin

PR #672, opened 2016-10-13
heppy updates for ICHEP from multileptons

peruzzim:
- soft muon ID, ICHEP version
- WP encoding modified for tau ID MVAOldDMRun2, dR=0.3 and 0.5


PhysicsTools/Heppy/python/analyzers/objects/TauAnalyzer.py
```diff
@@ -84,8 +87,8 @@ def id6(tau,X):
             tau.idMVA = id6(tau, "by%sIsolationMVArun2v1DBoldDMwLT")
             tau.idMVANewDM = id6(tau, "by%sIsolationMVArun2v1DBnewDMwLT")
             tau.idCI3hit = id3(tau, "by%sCombinedIsolationDeltaBetaCorr3Hits")
-            tau.idMVAOldDMRun2 = id3(tau, "by%sIsolationMVArun2v1DBoldDMwLT")
-            tau.idMVAOldDMRun2dR03 = id3(tau, "by%sIsolationMVArun2v1DBdR03oldDMwLT")
+            tau.idMVAOldDMRun2 = id5(tau, "by%sIsolationMVArun2v1DBoldDMwLT")
```
peruzzim: Warning: the encoding of tau.idMVAOldDMRun2 is not backward compatible (i.e. the cut value should be changed downstream to keep the same selection)

gpetruc:
@steggema: objections?

PhysicsTools/Heppy/python/analyzers/objects/TauAnalyzer.py
steggema: Not sure why this line was introduced - it's the same as line 87 but only saved 3 working points; now saves 5 working points. I would remove it

PhysicsTools/Heppy/python/analyzers/objects/TauAnalyzer.py
```diff
@@ -84,8 +87,8 @@ def id6(tau,X):
...
+            tau.idMVAOldDMRun2dR03 = id4(tau, "by%sIsolationMVArun2v1DBdR03oldDMwLT")
```
steggema: The deltaR 03 MVA isolation should also have 6 working points saved, so no need to introduce id4

PhysicsTools/Heppy/python/analyzers/objects/TauAnalyzer.py
peruzzim: Historical reasons, now it should be fixed (just new dR=0.3 ID added with 6 working points).

@gpetruc, please wait for merging, will require to adapt TauType in cmgtools.

PR #674, opened 2016-11-03
updates for W mass

perrozzi:


gpetruc:
I have no way of testing 5_X pull requests. 
Maybe it could be useful if you coordinate between WMass and the other SMP analyses for changes that are local to WMass.

PR #677, opened 2016-11-23
Heppy commits from susy multilepton

cheidegg:
* run dependent GT


PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
```diff
@@ -83,13 +109,23 @@ def __init__(self, cfg_ana, cfg_comp, looperName):
...
           elif doResidual not in [True,False]: raise RuntimeError, "If specified, applyL2L3Residual must be any of { True, False, 'MC', 'Data'(default)}"
-          GT = getattr(cfg_comp, 'jecGT', mcGT if self.cfg_comp.isMC else dataGT)
+          #GT = getattr(cfg_comp, 'jecGT', mcGT if self.cfg_comp.isMC else dataGT)
+          GTs = []
+          if self.cfg_comp.isMC:
+              GTs.append( getattr(cfg_comp, 'jecGT', mcGT) )
+              #print str(mcGT), getattr(cfg_comp, 'jecGT', mcGT)
+          else:
+              for gt in dataGT:
+                  GTs.append( getattr(cfg_comp, 'jecGT', gt) )
```
gpetruc: I think a more proper format for run-dependent JECs should be either a list of IOVs `((first run, last run,), GT)` or maybe better just `(first run, GT)` so that the last IOV is always open-ended.
Having two separate lists, one of GTs and one of runs, and relying on them being aligned by index, seems really error prone to me.

PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
```diff
@@ -218,11 +257,19 @@ def process(self, event):
                 self.jetsIdOnly.append(jet)
 
         jetsEtaCut = [j for j in self.jets if abs(j.eta()) <  self.cfg_ana.jetEta ]
-        self.cleanJetsAll, cleanLeptons = cleanJetsAndLeptons(jetsEtaCut, leptons, self.jetLepDR, self.jetLepArbitration)
-
+        self.cleanJetsAll=[]
+        cleanLeptons=[]
```
gpetruc: you don't really need these two lines of code for how the scoping of python works

PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
```diff
@@ -1,4 +1,5 @@
 import math, os
+import numpy as np
```
gpetruc: it doesn't seem like this is used anywhere

PhysicsTools/Heppy/python/analyzers/gen/LHEWeightAnalyzer.py
```diff
@@ -19,28 +19,49 @@ class LHEWeightAnalyzer( Analyzer ):
...
         super(LHEWeightAnalyzer,self).beginLoop(setup)
+        
+        if self.cfg_ana.useLumiInfo:
+            lumis = Lumis(self.cfg_comp.files)
+            for lumi in lumis:
+                if lumi.getByLabel('generator',self.genLumiHandle):
+                    weightNames = self.genLumiHandle.product().weightNames()
+                    for wn in weightNames:
+                        self.LHEWeightsNames.append(wn)
+                    break #so stupid to do it that way, crash otherwise
```
gpetruc: uh? can you explain better?

PhysicsTools/Heppy/python/analyzers/gen/LHEWeightAnalyzer.py
```diff
@@ -19,28 +19,49 @@ class LHEWeightAnalyzer( Analyzer ):
     def __init__(self, cfg_ana, cfg_comp, looperName ):
         super(LHEWeightAnalyzer,self).__init__(cfg_ana,cfg_comp,looperName)
 
+        self.LHEWeightsNames=[]
+        self.lumiCounter=-1
+        self.currentLumi=0
```
gpetruc: What are `self.lumiCounter` and `self.currentLumi` used for?
If, as it seems, they are not used for producing any of the output of the module then they should be removed.

gpetruc (changes requested):

gpetruc:
Also, please edit the PR description so that it lists all the features it adds (there's more than just run-dependent GTs here).
For Heppy PRs, it's important since we refer to our PRs to document the changes when we propagate them upstream to Colin and Andrea.

gpetruc:
@cheidegg @mmarionncern ping
(apologies to cbernet which I tagged by mistake...)

mmarionncern:
@gpetruc started to work on it, but won;t be able to finish before Wednesday next week

PhysicsTools/Heppy/python/analyzers/gen/LHEWeightAnalyzer.py
mmarionncern: Copying directly the list does not work, for a reason I don't understand. I had to copy the elements one by one to make it work. If there is a better option, i'd be happy to implement it

mmarionncern:
@gpetruc I made the requested changes in the code. Most of the items you wanted to address were remnant development stuff non needed anymore. Let me know if you see other actions to do.

PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
```diff
@@ -83,13 +105,19 @@ def __init__(self, cfg_ana, cfg_comp, looperName):
           if   doResidual == "MC":   doResidual = self.cfg_comp.isMC
           elif doResidual == "Data": doResidual = not self.cfg_comp.isMC
           elif doResidual not in [True,False]: raise RuntimeError, "If specified, applyL2L3Residual must be any of { True, False, 'MC', 'Data'(default)}"
-          GT = getattr(cfg_comp, 'jecGT', mcGT if self.cfg_comp.isMC else dataGT)
+          GTs = getattr(cfg_comp, 'jecGT', mcGT if self.cfg_comp.isMC else dataGT)
```
gpetruc, line 87: here for backwards compatibility you need something like
`if type(GTs) == str: GTs = [ (0, GT) ]`

PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
```diff
@@ -8,6 +8,28 @@
...
+      if abs(bM-bm)==1:
+	return bm
+         
+      if n>len(vals):
+          return -1
+      n+=1
+    
+    return -1;
+
+
```
gpetruc: all this is not needed, just add "from bisect import bisect"
https://docs.python.org/2.7/library/bisect.html

PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
```diff
@@ -83,13 +105,19 @@ def __init__(self, cfg_ana, cfg_comp, looperName):
...
                      'calculateType1METCorrection' :calculateType1METCorrection, }
           if kwargs['calculateType1METCorrection']: kwargs['type1METParams'] = cfg_ana.type1METParams
           # instantiate the jet re-calibrator
-          self.jetReCalibrator = JetReCalibrator(GT, cfg_ana.recalibrationType, doResidual, cfg_ana.jecPath, **kwargs)
+          self.jetReCalibrators={}
+          self.runsGT=[]
+          for iGT in range(0,len(GTs)):
+              self.jetReCalibrators[iGT] = JetReCalibrator(GTs[iGT][1], cfg_ana.recalibrationType, doResidual, cfg_ana.jecPath, **kwargs)
+              self.runsGT.append(GTs[iGT][0])
+          self.runsGT.append(1000000000)
```
gpetruc: some code cleanup:
 * make self.jetReCalibrators a list instead of a map, and append
 * you don't need the `self.runsGT.append(1000000000)` since `bisect` will return work even without it
 * you could do `for (run,GT) in GTs:` instead of iterating on the index, but that's not a big deal.

PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
```diff
@@ -126,6 +154,9 @@ def process(self, event):
         rho  = float(self.handles['rho'].product()[0])
         self.rho = rho
 
+        run=event.input.eventAuxiliary().id().run()
+        runBin=findBin(run, self.runsGT)
```
gpetruc: replace with `runBin = bisect(self.runsGT, run) - 1` (and throw and exception if the runBin is -1 at the end)

PhysicsTools/Heppy/python/analyzers/gen/LHEWeightAnalyzer.py
gpetruc: please put that explanation in the comments, so that if somebody has to modify that code in the future they will know ( `#so stupid to do it that way, crash otherwise` doesn't explain much )

gpetruc (changes requested):

PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py
mmarionncern: Done below in the code

gpetruc (approved):

gpetruc:
thanks. code looks fine to me. now testing.

PhysicsTools/Heppy/python/analyzers/gen/LHEWeightAnalyzer.py
```diff
@@ -19,28 +19,47 @@ class LHEWeightAnalyzer( Analyzer ):
...
+        self.LHEWeightsNames=[]
+
     def declareHandles(self):
         super(LHEWeightAnalyzer, self).declareHandles()
         self.mchandles['LHEweights'] = AutoHandle('externalLHEProducer',
                                                   'LHEEventProduct',
                                                   mayFail=True,
                                                   fallbackLabel='source',
                                                   lazy=False )
+        if self.cfg_ana.useLumiInfo:
```
steggema, line 31: I just realised this breaks existing code and that this won't run with the default config in the class since useLumiInfo isn't set. I would at least add the latter to the default config if this is an essential parameter. Otherwise you can also make it a facultative option by using getattr(self.cfg_ana, 'useLumiInfo', False).

PR #678, opened 2016-11-24
HeppyCore batchmanager: record jobs for which the submission failed

gpetruc:


gpetruc:
Closing since for the moment the submission_failed marker is giving false positives. will reopen (and merge) when fixed

PR #680, opened 2016-12-08
Update to latest 80X (CMSSW_8_0_24_patch1)

gpetruc:
This PR only updates the underlying CMSSW, it does not yet update the Heppy to the new version in https://github.com/cms-sw/cmssw/pull/16758
So, it should be transparent, but please help test it.
**The update to >= 8_0_20 is needed to correctly read the packed PF candidates from datasets produced in that release, i.e. the data re-reco and the Moriond MC (otherwise, variables like the number of hits will be broken)**


gpetruc:
@clelange @mschoene @mariadalfonso @steggema @aelwood @TaiSakuma @emanueledimarco 
if anyone can run a quick test today that setting up an area in `8_0_24_patch1` merging this doesn't break anything it would be nice, I would like to be able to merge this plus #681 by end of this week.

gpetruc:
After some testing, it appears `edmFileUtil --ls` segfaults on xrootd files in `CMSSW >= 8_0_21`  due to a xrootd update. So I'll close this PR and open one with `CMSSW_8_0_20`, which is sufficient to read re-reco data.

PR #681, opened 2016-12-08
Heppy: Spring16 electron MVA ID

peruzzim:
Merge of the CMSSW branch needed to calculate the HZZ and GP trainings of the Spring16 electron MVA ID (as in cms-sw/cmssw#16839 - thread to be watched to check if a rebase is decided before merging it here).

Also includes the addition of estimators and working points to the Heppy electron analyzer code.

Warning: the name of the estimators not anymore "Spring16", but rather "Spring16HZZ" and "Spring16GP". This will break existing code that calls Electron.mvaRun2("Spring16"). The HZZ version is the one equivalent to the previous behaviour.

Requires adding to the sparse-checkout file the following lines:
/RecoEgamma/ElectronIdentification/
/RecoEgamma/PhotonIdentification/


peruzzim:
To be noted: it also requires to include weight files in the external/ directory, following the POG recipe, as noted in CERN-PH-CMG/cmgtools-lite#155.

gpetruc:
Tested together with https://github.com/CERN-PH-CMG/cmgtools-lite/pull/155 in CMSSW_8_0_25, for which there is no need of getting the data files of the training, it's enough to add the two packages to the sparse checkout.
I will merge the two PRs, and update the twiki.

PR #682, opened 2016-12-09
Update to CMSSW 8_0_25

gpetruc:
The update to >= 8_0_20 is needed to correctly read the packed PF candidates from datasets produced in that release, i.e. the data re-reco and the Moriond MC (otherwise, variables like the number of hits will be broken)

There should be no changes in Heppy.

CMSSW releases after 8_0_20_patch1 all the way to 8_0_24_patch1 seem to have a broken combination of edmFileUtil + xrootd (+ root?) so they currently segfault when doing edmFileUtil on a file accessed via xrootd, so they are not recommended.
*The xrootd bug is fixed in 8_0_25, so this PR was updated to use that release.*

Replaces #680


clelange:
Tested VVResonances on an old signal sample. Seems to work fine, but output of the re-evaluation of the Hbb-tagger is different due to changes in the overall CMSSW b-tagging configuration (which are partly imported in the VVResonances part of heppy). Fine though to upgrade, since these changes need to be followed up elsewhere.

steggema:
Already tested H2TauTau with merge of 8_0_21 locally and works fine.

peruzzim:
CMSSW 8_0_25 has been released and does not show the edmFileUtil + xrootd problem (@clelange tested it). We could close this PR and open a new one with 8_0_25.

gpetruc:
Updated this PR to 8_0_25.

peruzzim (approved):
Multilep validation passed for 80X-MC and 80X-Data options.

PR #683, opened 2016-12-16
Heppy: fix missing default parameter in LHEWeightAnalyzer

peruzzim:
Bugfix for missing default parameter useLumiInfo in LHEWeightAnalyzer
Fixes https://github.com/CERN-PH-CMG/cmg-cmssw/pull/677#pullrequestreview-13307379

@mariadalfonso @steggema @simoneg90 please test

### DIFF
--- a/PhysicsTools/Heppy/python/analyzers/gen/LHEWeightAnalyzer.py
+++ b/PhysicsTools/Heppy/python/analyzers/gen/LHEWeightAnalyzer.py
@@ -93,5 +93,6 @@ class LHEWeightAnalyzer( Analyzer ):
 
 setattr(LHEWeightAnalyzer,"defaultConfig",
     cfg.Analyzer(LHEWeightAnalyzer,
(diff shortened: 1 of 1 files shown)


mariadalfonso:
@peruzzim @gpetruc
I tested this and works but there are other parameters in the jetAnalyzer that have the same problem
https://github.com/CERN-PH-CMG/cmg-cmssw/blob/heppy_80X/PhysicsTools/Heppy/python/analyzers/objects/JetAnalyzer.py#L495-L539

such as cleanJetsFromLeptons = True, and storeLowPtJets = False,
https://github.com/CERN-PH-CMG/cmgtools-lite/blob/80X/TTHAnalysis/python/analyzers/susyCore_modules_cff.py#L336
https://github.com/CERN-PH-CMG/cmgtools-lite/blob/80X/TTHAnalysis/python/analyzers/susyCore_modules_cff.py#L366

PR #686, opened 2017-01-30
Update jet ID implementation to latest recommendation

steggema:
This only affects jets with 2.7 < abs(eta) < 3.0.


gpetruc:
I think at some point we should version the IDs (e.g. having a `POG_PFID_Loose_2016`) so that we don't overwrite the old ones.

PR #688, opened 2017-01-30
Safe list passing in JetReCalibrator and get heppy help when writing "heppy --help"

steggema:


gpetruc:
What was the problem with the previous passing of lists?
(anyway, I'm ok with the fix)

steggema:
It's the usual issue with using lists as default arguments: http://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument
It actually appeared when I used the JetReCalibrator (I guess in a slightly different way than others did)

gpetruc:
On Mon, Jan 30, 2017 at 5:56 PM, Jan Steggemann <notifications@github.com>
wrote:

> It's the usual issue with using lists as default arguments:
> http://stackoverflow.com/questions/1132941/least-astonishmen
> t-and-the-mutable-default-argument
> It actually appeared when I used the JetReCalibrator (I guess in a
> slightly different way than others did)
>
I'm aware of that python feature, but I didn't realize that it would break
correctAll.

PR #690, opened 2017-02-13
Update spring16 electron mvaID WP

GaelTouquet:
Update spring16 electron mvaID WP


gpetruc:
Hi @GaelTouquet,

I see that the WP80 and WP90 numbers match don't match exactly the ones in https://github.com/ikrav/cmssw/blob/egm_id_80X_v2/RecoEgamma/ElectronIdentification/python/Identification/mvaElectronID_Spring16_GeneralPurpose_V1_cff.py#L46-L63 : the twiki has numbers rounded to 3 decimals, the cff has not, and heppy conforms to the twiki but not to the cff.
This can cause issues in synchronizations, as seen by @jheikkil

Opinions on what do do about it?

PR #691, opened 2017-02-14
Heppy updates to json analyzer and heppy_add

rmanzoni:
1. add the possibility to check a second JSON file
   1. can specify a second JSON file to a component
   2. events will still be skimmed according to the main JSON
   3. for each event, an attribute is saved as to whether it is contained in the secondary JSON or not
   4. I found this useful for some trigger studies

2. heppy_hadd.py: allow to directly hadd files stored on EOS
   1. hadd of trees stored on EOS
   2. read their location from tree.root.url
   3. eliminated the need of copying everything to afs just to hadd trees (that implied ~2x tree size to be available on afs)

### DIFF
--- a/PhysicsTools/Heppy/python/analyzers/core/JSONAnalyzer.py
+++ b/PhysicsTools/Heppy/python/analyzers/core/JSONAnalyzer.py
@@ -37,6 +37,12 @@ class JSONAnalyzer( Analyzer ):
         else:
             self.lumiList = None
         
+        if hasattr(self.cfg_comp, 'additionaljson'):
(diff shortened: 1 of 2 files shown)


gpetruc:
@peruzzim @clelange @steggema can one of you test the heppy_add change and merge?
I'm trying to prepare a branch for 9XY, and the first step is to get a 80X branch with the latest upstream heppy (since that one is also in the 9XY release) and I would like to have all the heppy_80X PRs closed for that.

cbernet:
Hi all, now, it's possible to access eos as a standard filesystem as you probably know. It's mounted on lxplus, for example. So is there really a need for specific code? I don't know, I haven't tried.

gpetruc (approved):

PR #692, opened 2017-02-14
added muonID method for run-dependent ID

GaelTouquet:
added muonID method for run-dependent ID


gpetruc (changes requested):
Please change this to be another case into the `muonID()` function, just like all other IDs?

GaelTouquet:
This new commit should fix the issue

gpetruc (approved):

PR #703, opened 2017-07-24
Heppy: use local EOS client instead of the AFS-based one

peruzzim:


gpetruc (approved):
I suppose we could merge this even if we should also do a cleanup of the duplication between heppy and cmgtools for all this stuff

PR #704, opened 2017-08-17
Heppy: METanalyzer, gainSwitch

mariadalfonso:


PhysicsTools/Heppy/python/analyzers/objects/METAnalyzer.py
```diff
@@ -66,9 +66,37 @@ def makeTkMETs(self, event):
...
+        doPuppiloose=getattr(self.cfg_ana,"includeTkMetPuppiPVLoose",True)
+        doPuppiPVUsedInFit=getattr(self.cfg_ana,"includeTkMetPuppiPVUsedInFit",True)
+        doPuppiNoPV=getattr(self.cfg_ana,"includeTkMetPuppiNoPV",True)
+        doPuppiUsedInFitTight=getattr(self.cfg_ana,"includeTkMetPuppiUsedInFitTight",True)
+        doPuppiUsedInFitLoose=getattr(self.cfg_ana,"includeTkMetPuppiUsedInFitLoose",True)
+        doPuppiCompatibilityDz=getattr(self.cfg_ana,"includeTkMetPuppiCompatibilityDz",True)
+        doPuppiCompatibilityBTag=getattr(self.cfg_ana,"includeTkMetPuppiCompatibilityBTag",True)
+        doPuppiNotReconstructedPrimary=getattr(self.cfg_ana,"includeTkMetPuppiNotReconstructedPrimary",True)
+        doPuppiOtherDeltaZ=getattr(self.cfg_ana,"includeTkMetPuppiOtherDeltaZ",True)
+
```
gpetruc: Some of these are also in the default config.
Also, the default config is false, but the default in the code is true.
I would leave the default to false also in the code, as I think they're mostly for MET commissioning, the average user won't want to run all them, right?

PhysicsTools/Heppy/python/analyzers/objects/autophobj.py
```diff
@@ -53,6 +53,8 @@
...
     NTupleVariable("eleCutIdCSA14_50ns_v1",     lambda x : (1*x.electronID("POG_Cuts_ID_CSA14_50ns_v1_Veto") + 1*x.electronID("POG_Cuts_ID_CSA14_50ns_v1_Loose") + 1*x.electronID("POG_Cuts_ID_CSA14_50ns_v1_Medium") + 1*x.electronID("POG_Cuts_ID_CSA14_50ns_v1_Tight")) if abs(x.pdgId()) == 11 else -1, int, help="Electron cut-based id (POG CSA14_50ns_v1): 0=none, 1=veto, 2=loose, 3=medium, 4=tight"),
     NTupleVariable("eleCutIdSpring15_25ns_v1",     lambda x : (1*x.electronID("cutBasedElectronID-Spring15-25ns-V1-standalone-veto") + 1*x.electronID("cutBasedElectronID-Spring15-25ns-V1-standalone-loose") + 1*x.electronID("cutBasedElectronID-Spring15-25ns-V1-standalone-medium") + 1*x.electronID("cutBasedElectronID-Spring15-25ns-V1-standalone-tight")) if abs(x.pdgId()) == 11 and x.isElectronIDAvailable("cutBasedElectronID-Spring15-25ns-V1-standalone-veto") else -1, int, help="Electron cut-based id (POG Spring15_25ns_v1): 0=none, 1=veto, 2=loose, 3=medium, 4=tight"),
+    # more on electron
+    NTupleVariable("hasGainSwitchFlag",   lambda x : x.userInt("hasGainSwitchFlag") if hasattr(x,'hasGainSwitchFlag') and abs(x.pdgId()) == 11 else -1, int, help="photon has gain switched"),
```
gpetruc: it should be `if x.hasUserFloat('...')`, not `if hasattr(x, '...')`

PhysicsTools/Heppy/python/analyzers/objects/autophobj.py
```diff
@@ -180,6 +182,7 @@
     NTupleVariable("relIso", lambda x : x.ftprRelIso03 if hasattr(x,'ftprRelIso03') else x.relIso, float, help="relativeIsolation for photons with footprint removal and pile-up correction"),
     NTupleVariable("mcMatchId",  lambda x : getattr(x, 'mcMatchId', -99), int, mcOnly=True, help="Match to source from hard scatter (pdgId of heaviest particle in chain, 25 for H, 6 for t, 23/24 for W/Z), zero if non-prompt or fake"),
     NTupleVariable("mcPt",   lambda x : x.mcGamma.pt() if getattr(x,"mcGamma",None) else 0., mcOnly=True, help="p_{T} of associated gen photon"),
+    NTupleVariable("hasGainSwitchFlag", lambda x: x.userInt("hasGainSwitchFlag") if hasattr(x,'hasGainSwitchFlag') else -1, int, help="photon has gain switched"),
```
gpetruc: as above

PhysicsTools/Heppy/python/analyzers/objects/METAnalyzer.py
```diff
@@ -85,23 +113,70 @@ def makeTkMETs(self, event):
...
+                if doPVUsedInFit and pvflag== 3:
+                    chargedPVUsedInFit.append(pxy)
+
+                #adding pvAssociationQuality stuff
+                pvAssflag = pfcand.pvAssociationQuality()
+
+                if doUsedInFitTight and pvAssflag == 7 :
+                    chargedUsedInFitTight.append(pxy)
+
+                if doUsedInFitLoose and pvAssflag == 6 :
```
gpetruc: wouldn't it make more sense to do it with `>=` instead of `==` ?
also, shouldn't you also check `pfRef.key()` ? the association quality only tells you how well it's associated to its PV, it doesn't tell you whether it's PV[0] or another one.

gpetruc (changes requested):

mariadalfonso:
@gpetruc fixed the METanalyze and the useInt

gpetruc (approved):

PR #705, opened 2017-08-25
Use precise spring16 electron raw MVA values instead of the rounded ones given in the twiki

jheikkil:
Update the spring16 electron raw MVA values to correspond the ones given here: https://github.com/ikrav/cmssw/blob/egm_id_80X_v2/RecoEgamma/ElectronIdentification/python/Identification/mvaElectronID_Spring16_GeneralPurpose_V1_cff.py#L48-L63

The values given in the egamma pog twiki are rounded. This can cause issues e.g. when synchronizing.
See also: https://github.com/CERN-PH-CMG/cmg-cmssw/pull/690#issuecomment-320628496

### DIFF
--- a/PhysicsTools/Heppy/python/physicsobjects/Electron.py
+++ b/PhysicsTools/Heppy/python/physicsobjects/Electron.py
@@ -70,8 +70,8 @@ class Electron( Lepton ):
             'dxy' : abs(self.dxy()),
             'dz' : abs(self.dz()),
             'chi2' : self.normalizedGsfChi2(),
-            'ECALPFIsoEA' : self.hltPFIso('ECALPFIsoEA')/self.pt(),
-            'HCALPFIsoEA' : self.hltPFIso('HCALPFIsoEA')/self.pt(),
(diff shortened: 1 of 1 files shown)


PhysicsTools/Heppy/python/physicsobjects/Electron.py
```diff
@@ -70,8 +70,8 @@ def cutBasedId(self, wp, showerShapes="auto"):
             'dxy' : abs(self.dxy()),
             'dz' : abs(self.dz()),
             'chi2' : self.normalizedGsfChi2(),
-            'ECALPFIsoEA' : self.hltPFIso('ECALPFIsoEA')/self.pt(),
-            'HCALPFIsoEA' : self.hltPFIso('HCALPFIsoEA')/self.pt(),
+            'ECALPFIsoEA' : self.hltPFIso('ECALPFIsoEA')/self.pt() if self.rhoHLT != None in wp else None,
+            'HCALPFIsoEA' : self.hltPFIso('HCALPFIsoEA')/self.pt() if self.rhoHLT != None in wp else None,
```
gpetruc: I think it should it be ` if "HLT" in wp else None`

gpetruc (changes requested):

gpetruc (approved):

PR #706, opened 2017-10-03
Addition of annealus lepton isolations

mmarionncern:
Compute the annealus isolations for each lepton.

### DIFF
--- a/PhysicsTools/Heppy/python/analyzers/objects/LeptonAnalyzer.py
+++ b/PhysicsTools/Heppy/python/analyzers/objects/LeptonAnalyzer.py
@@ -200,6 +200,7 @@ class LeptonAnalyzer( Analyzer ):
         if self.doIsoAnnulus:
             for lep in event.inclusiveLeptons:
                 self.attachIsoAnnulus04(lep)
+                self.attachIsoAnnealusMiniIso(lep)
 
         if self.doIsolationScan:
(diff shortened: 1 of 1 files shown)


PhysicsTools/Heppy/python/analyzers/objects/LeptonAnalyzer.py
```diff
@@ -200,6 +200,7 @@ def makeLeptons(self, event):
         if self.doIsoAnnulus:
             for lep in event.inclusiveLeptons:
                 self.attachIsoAnnulus04(lep)
+                self.attachIsoAnnealusMiniIso(lep)
```
gpetruc, line 203: why one is Annulus and the other Annealus?
could we use just one spelling, whatever is more correct

gpetruc (approved):
looks ok to me for the rest

gpetruc:
@mmarionncern does this need to be forward-ported to 94X?

PhysicsTools/Heppy/python/analyzers/objects/LeptonAnalyzer.py
mmarionncern: Yes, going to correct that asap in the different releases

mmarionncern:
@gpetruc yes, it would be good to have it as well in 94X

PR #707, opened 2017-11-29
WIP: adding extra variables for W mass analysis

pfs:
I have made some changes in this analyzer, in the context of the W mass analysis.
If I haven't overlooked anything in principle they shouldn't break anything wrt to the previous version, but i also didn't test thoroughly.  If there are some test cases i would be happy to try them out.


gpetruc:
Hi,
This doesn't merge anymore after merging #704 
Can you have a look?

pfs:
No problem. I'll cancel this one and make a new PR as i there were some more changes in between as well

PR #710, opened 2017-12-18
Heppy 80x wptdev

pfs:
This supersedes #707 and merges with the changes introduced by @mariadalfonso in #704 
@emanueledimarco @mdunser follow this as well


gpetruc:
Looks ok to me. Is there anybody outside of WMass that is using the track quantities that wants to check if this PR affects their code?
If I get no reply, I'll merge it on thursday afternoon on all branches (80X, 92X, 94X)

PR #713, opened 2018-01-12
Condor submission

shane-breeze:
See https://github.com/CERN-PH-CMG/cmgtools-lite/pull/206


gpetruc:
At cern the condor system has access to afs, so we may go without the tarring and transferring of the inputs (OTOH we know that the current system reading from AFS is a bit hard on the servers when submitting thousands of jobs, so perhaps transferring the inputs via condor may be good anyway)

PR #727, opened 2018-03-29
Minor fix to 2017 electron MVA evaluation, plus add relevant E/Gamma code

steggema:
The E/Gamma developments are also merged into CMSSW (https://github.com/cms-sw/cmssw/pull/21838/ master,  https://github.com/cms-sw/cmssw/pull/22303 CMSSW_9_4_MAOD_X), but I'm not sure they'll get into central CMSSW_9_4_X, so it probably makes sense to merge them directly.


peruzzim:
Will be mostly replaced by #730, apart from the small include fix.

steggema:
I think it should merge cleanly with only the import fix left once #730 is merged (so let's wait until then), otherwise we can have a separate PR.

PR #731, opened 2018-05-22
Bug fix to MET analyzer, adding 2017 cut-based ID for electrons

avartak:
There is a bug in the METAnalyzer in the way in which metNoMuNoPU and metNoEleNoPU are handled. This is fixed in this PR.

I have also added the 2017 cut-based ID for electrons which seems to have been missing so far.

### DIFF
--- a/PhysicsTools/Heppy/python/physicsobjects/Electron.py
+++ b/PhysicsTools/Heppy/python/physicsobjects/Electron.py
@@ -66,11 +66,17 @@ class Electron( Lepton ):
             wp = wp.replace("_zs","")
         elif showerShapes == "auto":
             showerShapes = "full5x5"
+
+        hOverE_CE, hOverE_Cr = 0, 0
+        if "POG_FALL17_94X_v1" in wp:
+            hOverE_CE = 1.12 if self.physObj.isEB() else 0.5
+            hOverE_Cr = 0.0368 if self.physObj.isEB() else 0.201
+
         vars = {
             'dEtaIn' : abs(self.dEtaInSeed()) if (("POG_SPRING16_25ns_v1" in wp) or ("POG_FALL17_94X_v1" in wp)) else abs(self.physObj.deltaEtaSuperClusterTrackAtVtx()),
             'dPhiIn' : abs(self.physObj.deltaPhiSuperClusterTrackAtVtx()),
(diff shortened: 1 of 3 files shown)


PhysicsTools/Heppy/python/physicsobjects/Electron.py
```diff
@@ -67,10 +67,10 @@ def cutBasedId(self, wp, showerShapes="auto"):
         elif showerShapes == "auto":
             showerShapes = "full5x5"
         vars = {
-            'dEtaIn' : abs(self.dEtaInSeed()) if "POG_SPRING16_25ns_v1" in wp else abs(self.physObj.deltaEtaSuperClusterTrackAtVtx()),
+            'dEtaIn' : abs(self.dEtaInSeed()) if (("POG_SPRING16_25ns_v1" in wp) or ("POG_FALL17_94X_v1" in wp)) else abs(self.physObj.deltaEtaSuperClusterTrackAtVtx()),
             'dPhiIn' : abs(self.physObj.deltaPhiSuperClusterTrackAtVtx()),
             'sigmaIEtaIEta' : self.physObj.full5x5_sigmaIetaIeta() if showerShapes == "full5x5" else self.physObj.sigmaIetaIeta(),
-            'H/E' : self.physObj.hadronicOverEm(),
+            'H/E' : self.physObj.hadronicOverEm() - 1.12/(self.physObj.superCluster().energy()) - 0.201*self.rho/(self.physObj.superCluster().energy()) if "POG_FALL17_94X_v1" in wp else self.physObj.hadronicOverEm(),
```
gpetruc: from the twiki it looks like the C_0 definition is different for barrel and endcaps.

also, this line is really too long to be readable. I would suggest to put before the map a block like
````
  hOverE_CE, hOverE_Cr = 0, 0
  if "POG_FALL17_94X_v1" in wp:
      hOverE_CE = 1.12 if self.physObj.isEB() else 0.5
      hOverE_Cr = 0.0368 if self.physObj.isEB() else 0.201
````
and then you can have just
````
  'H/E' : self.physObj.hadronicOverEm() - (hOverE_CE  + hOverE_Cr * self.rho)/(self.physObj.superCluster().energy())
````

gpetruc (changes requested):
H/E cut needs fixes. the rest looks ok to me

avartak:
Ok, pushed the changes you suggested

gpetruc (approved):

PR #736, opened 2018-08-15
Heppy updates: 2017 tau variables and features sync to nanoAOD

peruzzim:
Summary of changes:

- 2017 tau ID variables and MVA score evaluation by @steggema and @gpetruc (includes changes to RecoTauTag/RecoTau)
- taus: always use dxy and dz from leadChargedHadrCand, ignoring vertex choice, as in nanoAOD
- added SUSY working points to Fall17 electron ID
- JetAnalyzer: option to match leptons and taus to jets using candidatePtr instead of deltaR, as in nanoAOD
- LHE content: option for storage of PS weights and MG reweighting factors with new syntax


gpetruc:
Looks ok to me. 
@steggema any opinion on taus?

steggema:
#737 has a few more updates that will also allow to evaluate the MVA ID working points, but this should merge graciously since the #737 is based on the developments/commits included here so I think the MVA ID part is OK. 

The other change is certainly fine if the goal is to have the same information as in NanoAOD - one would typically still apply an additional cut to make sure that the tau vertex corresponds to the primary vertex, and I'm not sure how this is solved in the TTH code.

peruzzim:
This code is to get the value of the dxy with respect to the PV, a selection is later required on this value as a part of the standard tau ID requirements.

steggema (approved):
The changes look good to me.

PR #738, opened 2018-08-18
Heppy: streamlined isolation interface, enabling the use of electron EA iso with cone 0.3

GaelTouquet:
To implement 2017 EA electron isolation properly, we realized that it was necessary to clean up the lepton isolation interface. Slight changes in client code will be necessary this time but we think that the new interface will be able to accommodate easily future implementations of the isolation.

The isolation has been perfectly synchronized with other groups in the h to tau tau analysis.

### DIFF
--- a/PhysicsTools/Heppy/python/physicsobjects/Electron.py
+++ b/PhysicsTools/Heppy/python/physicsobjects/Electron.py
@@ -5,6 +5,25 @@ import ROOT
 import sys
 from math import exp
 
+effective_areas = dict(
+    ea03 = [
+        (1.000, 0.1440),
+        (1.479, 0.1562),
+        (2.000, 0.1032),
+        (2.200, 0.0859),
+        (2.300, 0.1116),
+        (2.400, 0.1321),
+        (2.500, 0.1654)        
+        ]
+)
+
+def get_effective_area(eta, area):
+    eta = abs(eta)
+    ea_defs = effective_areas[area]
+    for etamax, ea in ea_defs: 
+        if eta < etamax:
+            return ea 
+    return ea_defs[-1][1]
 
 class Electron( Lepton ):
 
@@ -24,6 +43,8 @@ class Electron( Lepton ):
         self._mvaTrigV0     = {True:None, False:None}
         self._mvaTrigNoIPV0 = {True:None, False:None}
         self._mvaRun2 = {}
+        self.effarea_03 = get_effective_area(self.physObj.superCluster().eta(),
+                                             'ea03')
 
     def electronID( self, id, vertex=None, rho=None ):
         if id is None or id == "": return True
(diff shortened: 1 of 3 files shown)


PhysicsTools/Heppy/python/physicsobjects/Electron.py
```diff
@@ -5,6 +5,25 @@
...
+    ea03 = [
+        (1.000, 0.1440),
+        (1.479, 0.1562),
+        (2.000, 0.1032),
+        (2.200, 0.0859),
+        (2.300, 0.1116),
+        (2.400, 0.1321),
+        (2.500, 0.1654)        
+        ]
+)
```
gpetruc, line 18: for EA's, they're already in the LeptonAnalyzer, that also has also a proper switch for the different EA values depending on the MC era they're based on (e.g. spring16 vs summer16 vs fall17): https://github.com/CERN-PH-CMG/cmg-cmssw/blob/heppy_94X_dev/PhysicsTools/Heppy/python/analyzers/objects/LeptonAnalyzer.py#L361
I don't necessarily object at refactoring them to the electron object (or maybe to another file loaded by the electron object), but I *do* object at having them here as a single set of numbers and with no era and no reference to the source.

PhysicsTools/Heppy/python/physicsobjects/Electron.py
```diff
@@ -5,6 +5,25 @@
...
+
+def get_effective_area(eta, area):
+    eta = abs(eta)
+    ea_defs = effective_areas[area]
+    for etamax, ea in ea_defs: 
+        if eta < etamax:
+            return ea 
+    return ea_defs[-1][1]
```
gpetruc, line 26: also, I think that this and the above should be hidden from `import *` by having names starting with `_`, since they're not part of the public interface of the electron object and the name is unspecific enough that may clash with something else.

cbernet:
Hi, no problem I can deal with that on Monday when I can get my hands on a computer.



Le 21 août 2018 15:03:49 GMT+02:00, Giovanni Petrucciani <notifications@github.com> a écrit :
>gpetruc commented on this pull request.
>
>
>
>> +        (1.479, 0.1562),
>+        (2.000, 0.1032),
>+        (2.200, 0.0859),
>+        (2.300, 0.1116),
>+        (2.400, 0.1321),
>+        (2.500, 0.1654)        
>+        ]
>+)
>+
>+def get_effective_area(eta, area):
>+    eta = abs(eta)
>+    ea_defs = effective_areas[area]
>+    for etamax, ea in ea_defs: 
>+        if eta < etamax:
>+            return ea 
>+    return ea_defs[-1][1]
>
>also, I think that this and the above should be hidden from `import *`
>by having names starting with `_`, since they're not part of the public
>interface of the electron object and the name is unspecific enough that
>may clash with something else.
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/CERN-PH-CMG/cmg-cmssw/pull/738#pullrequestreview-148047717

Colin
-- sent from my phone

cbernet:
@gpetruc I fully agree with your comments (and I had missed the LeptonAnalyzer because I do not use it). I am not completely sure how to proceed, so let me explain my point of view:
- I think that the isolation part of [Lepton.py](https://github.com/CERN-PH-CMG/cmg-cmssw/blob/heppy_94X_dev/PhysicsTools/Heppy/python/physicsobjects/Lepton.py) has evolved to be rather cumbersome. To solve this I propose two methods that would be used in all cases: absIso and relIso. All R methods would be removed and R would be provided as an optional argument. 
- I got wary of monolithic analyzers that do more than what is needed by the user, and that's why I think one should be able to ask the electron for its isolation directly. 

That being said, again, you're completely right, we need a clean way to deal with eras. 

What I propose is to introduce a global configuration module for heppy (that's what I did on the FCC side for other reasons: https://github.com/cbernet/heppy/blob/master/configuration.py
Based on this configuration, decisions are taken, e.g. here: https://github.com/cbernet/heppy/blob/master/utils/deltar.py#L14

We could do the same in CMS. The user could then specify the era in the config file like this: 
https://github.com/cbernet/fcc-ee-higgs/blob/master/analysis_ee_ZH_lltautau.py#L48

If you agree with this way to proceed, I will: 
- introduce a configuration module containing only the eras for now (the ones in LeptonAnalyzer) 
- introduce a new module for EA definition
- load in the electron the correct EAs based on the era in Electron.py, and remove this part from LeptonAnalyzer. 

Please let me know what you think.

gpetruc:
Hi,
My proposal would be to take the EA setting from: (1) named function argument of the isolation function, (2) datamember of the electron object (3) data member of the electron class, and raise exception if neither of the 3 is available. 
The lepton analyzer can set it on the objects and then rely on (2)

cbernet:
perfect, working on it.

cbernet:
can be closed, overwritten by PR #741

PR #745, opened 2019-01-12
Enabling  heppy on Fermilab Condor cluster

bachtis:
The FNAL LPC condor has slightly different  setup than the CERN one .
This PR makes possible to run heppy with file transfer at the LPC .
It is fully transparent to everything else.

As I was testing it , the remote version crashed due to a mistake on the Lepton class.
what is interesting is that locally, python does not crash the job for this.
  
The printout was printing relIso() but in the current version relIso needs 2 arguments so I just
removed the printout


bachtis:
I assume I can merge this right?

PR #737, opened 2018-08-17
rerunning MVA tau isolation in heppy

GaelTouquet:
- cross-checked to have the same isolation scores and WP as following the CMSSW recipe given here https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuidePFTauID#Rerunning_of_the_tau_ID_on_M_AN1 .
- isolation perfectly synchronised with other groups for the 2017 HTT anaylsis.

PR #748, opened 2019-04-11
Fix dr03 tau MVA

gpetruc:
Fix regression from https://github.com/CERN-PH-CMG/cmg-cmssw/pull/737/files#diff-db9724459059ef09a84499d3a628e9d8 :  the `puCorrPtSumdR03` does not exist n 2017 MiniAOD v2., only `puCorrPtSum` is available.

Verified that with this change I get the right MVA output as in NanoAODv4 (`CMSSW_10_2_9`, era `Run2_2017,run2_nanoAOD_94XMiniAODv2`)

@GaelTouquet @steggema @peruzzim


GaelTouquet:
Indeed it seems that this change was tested and dropped as we didn't keep it in our analysis repository.
I must have cherry-picked too much preparing that commit. So very sorry!

PR #749, opened 2019-04-12
2017 Electron ID v2

GaelTouquet:
Most of the changes come from merging code from Egamma, as can be seen in [this issue](https://github.com/cms-sw/cmssw/issues/23676) and [this merged pull request](https://github.com/cms-sw/cmssw/pull/25337).
Most changes from us are in PhysicsTools/Heppy .
Also it seems some tau changes have slipped from [this commit](https://github.com/CERN-PH-CMG/cmg-cmssw/pull/737) about the TauID (although I don't understand why they are shown here even though the PR was merged...).
Feel free to review and suggest changes.

@cbernet @steggema @lucastorterotot


PhysicsTools/Heppy/python/physicsutils/EffectiveAreas.py
```diff
@@ -86,23 +86,23 @@
...
+        # https://github.com/cms-sw/cmssw/blob/master/RecoEgamma/ElectronIdentification/data/Fall17/effAreaElectrons_cone03_pfNeuHadronsAndPhotons_94X.txt
         'electron' : { '03' : 
-                       [ (1.000, 0.1566),
-                         (1.479, 0.1626),
-                         (2.000, 0.1073),
-                         (2.200, 0.0854),
-                         (2.300, 0.1051),
-                         (2.400, 0.1204),
-                         (2.500, 0.1524) ],
+                       [ (1.000, 0.1440),
```
gpetruc, line 91: This is the change Fall17  v1 -> Fall 17 v2?
If yes, I think it would be best to define a new set Fall17v2 and rename the current one as Fall17v1 (so that people are forced to take a decision)

gpetruc (changes requested):
You're deleting a lot of IDs in Electron.py that don't have a VID-equivalent but are used in other analysis.

PhysicsTools/Heppy/python/physicsutils/EffectiveAreas.py
cbernet: Yes, it is Fall 17 v2. @GaelTouquet can certainly add back Fall17v1.

cbernet:
> You're deleting a lot of IDs in Electron.py that don't have a VID-equivalent but are used in other analysis.

Yes, I had to do a major clean up of the infrastructure, there was a lot of obsolete / illogical code in this class. I think I've added all the IDs supported by e/gamma. 
I guess the new infrastructure can accommodate the other IDs. But since we're not using them, it would be difficult for us to implement them.

peruzzim:
@cbernet actually I'm afraid that many of those ID are needed, as they are used e.g. in SUSY or in ongoing versions of other analyses that have not updated to the latest ID versions

cbernet:
@peruzzim Yes, no problem, I'm sure they can be re-added easily in the new infrastructure. But as I said, it would be difficult for us to do it as we are not using these ids. Maybe you guys can survey which IDs are still needed and re-add them after the merge? I will be  happy to assist and explain how the new infrastructure is working. 

The reason why I had to do all this is that we could not re-code the v2 electron ID for heppy, as was done in the past. Instead, @guitargeek and myself teamed up to provide official egamma tools to run the ID in FWLite. So now, we simply use the e/gamma cmssw tools in FWLite. This makes it much easier for us and, in the future, we will get the new IDs almost for free.

Getting v2 to work was actually quite complicated, we spent maybe a month on this. 

To make it work, we had to redesign the ID interface in Heppy, because it was really chaotic. I guess that it evolved over the years without any real code review. So we took the easy path to remove the IDs that were not supported by e/gamma, as I thought that no analysis would be approved with an unsupported ID anyway. Also, we thought that this fork was kind of dead, so we decided to just maintain our own version. This development was already picked up by @bachtis from our fork, as far as I know. 

Now the electron ID is obtained through a single easy method, using normalized names, like this: 

https://github.com/cbernet/cmgtools-lite/blob/htt_9_4_11_cand1_v1/H2TauTau/python/heppy/sequence/common.py#L163

gpetruc:
I think the only way forward at the moment is to keep both the old (by hand) and new (VID-based) backends for the id.
There's some parts of the old implementation (all run1-related IDs) that can be safely dropped, but many run2 ones exist and don't necessarily have the VID implementation (e.g. even just for the cut-based IDs, VID doesn't provide also the working points for ID-only without also the isolation cut).

cbernet:
Yes, I think it’s a good solution.

On 12 Apr 2019, at 15:06, Giovanni Petrucciani <notifications@github.com<mailto:notifications@github.com>> wrote:


I think the only way forward at the moment is to keep both the old (by hand) and new (VID-based) backends for the id.
There's some parts of the old implementation (all run1-related IDs) that can be safely dropped, but many run2 ones exist and don't necessarily have the VID implementation (e.g. even just for the cut-based IDs, VID doesn't provide also the working points for ID-only without also the isolation cut).

PhysicsTools/Heppy/python/physicsobjects/Electron.py
```diff
@@ -19,579 +23,117 @@ def _physObjInit(self):
...
+        self._mvaid_normscore = {}
+        self._mvaid_passed = {}
+
+    def mva_category(self, name):
+        '''takes an mva name and returns the category of the electron,
+        only available using FWLite and Egamma code.'''
+        if 'cutBased' in name :
+            return None
+        if name not in self._mvaid_category :
+            category = self._mva_score_and_category(name)[1]
```
gpetruc, line 37: shouldn't this be `self._mvaid_category[name] = self._mva_score_and_category(name)[1]` ?

PhysicsTools/Heppy/python/physicsobjects/Electron.py
```diff
@@ -19,579 +23,117 @@ def _physObjInit(self):
...
+
+    def mva_score(self, name, norm=False):
+        '''returns the score of the given mva,
+        only available using FWLite and Egamma code.'''
+        if 'cutBased' in name :
+            return None
+        if norm :
+            if name not in self._mvaid_normscore :
+                score_raw = self._mva_score_and_category(name)[0]
+            return self._mvaid_normscore[name]
```
gpetruc, line 52: again `score_raw  = ...` should probably be `self._mvaid_normscore[name] = ...`

PhysicsTools/Heppy/python/physicsobjects/Electron.py
```diff
@@ -19,579 +23,117 @@ def _physObjInit(self):
...
+        else:
+            if name not in self._mvaid_score :
+                score_raw = self._mva_score_and_category(name)[0]
```
gpetruc, line 55: ditto

gpetruc (changes requested):
I'm having a look myself at how to merge this with the old implementation for the IDs that are still used, but please confirm that the lines of code I pointed out in this last review message should be fixed as per their comment.

PhysicsTools/Heppy/python/physicsobjects/Electron.py
lucastorterotot: Hi Giovanni,
as you can see in [the definition of `_mva_score_and_category`](https://github.com/CERN-PH-CMG/cmg-cmssw/pull/749/files/bd7006c265f72c3c55736a1617e9c7f3c1952db8#diff-4f88e01b20d7d82ca6444ccd2406453cR117), this method fills three dictionaries:
- `_mvaid_category`
- `_mvaid_score`
- `_mvaid_normscore`

So your proposal to write `self._mvaid_category[name] = self._mva_score_and_category(name)[1]` would lead to redefine `self._mvaid_category[name]` with its own previously assigned value.
Thought this code is not easily understandable, it provides the requested values without re-computing them if already computed.

Of course one could design this better. What do you think?

PhysicsTools/Heppy/python/physicsobjects/Electron.py
lucastorterotot: See [this comment](https://github.com/CERN-PH-CMG/cmg-cmssw/pull/749/files/bd7006c265f72c3c55736a1617e9c7f3c1952db8#r275395718).

PhysicsTools/Heppy/python/physicsobjects/Electron.py
lucastorterotot: See [this comment](https://github.com/CERN-PH-CMG/cmg-cmssw/pull/749/files/bd7006c265f72c3c55736a1617e9c7f3c1952db8#r275395718).

PR #750, opened 2019-04-12
Jet Energy Scale systematics can be split by sources and grouped

GaelTouquet:
For the H2TauTau analysis, JES is split by sources and grouped as is explained [here](https://twiki.cern.ch/twiki/bin/viewauth/CMS/HiggsToTauTauWorking2017#JES_uncertainty).
I have implemented the possibility of doing so in the jet recalibrator. Maybe this could be useful in other analysis too. 
This has not been fully validated yet but will be within a week, I will add a comment here when it will be so.

### DIFF
--- a/PhysicsTools/Heppy/python/physicsutils/JetReCalibrator.py
+++ b/PhysicsTools/Heppy/python/physicsutils/JetReCalibrator.py
@@ -148,8 +148,8 @@ class JetReCalibrator:
                 for group, sources in self.groupForUncertaintySources.iteritems():
                     shift_up = self.getCorrection(jet,rho,delta+1.0,sources=sources)
                     shift_down = self.getCorrection(jet,rho,delta-1.0,sources=sources)
-                    setattr(jet, "corr_"+group+"_JECup", shift_up)
-                    setattr(jet, "corr_"+group+"_JECdown", shift_down)
+                    setattr(jet, "corr_"+group+"_JEC_up", shift_up)
+                    setattr(jet, "corr_"+group+"_JEC_down", shift_down)
         if corr <= 0:
             return False
         newpt = jet.pt()*raw*corr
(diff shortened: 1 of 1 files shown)


PhysicsTools/Heppy/python/physicsutils/JetReCalibrator.py
```diff
@@ -67,16 +75,27 @@ def getCorrection(self,jet,rho,delta=0,corrector=None):
...
-            corr *= max(0, 1+delta*jet.jetEnergyCorrUncertainty)
+            if sources:
+                group_shift = 0.
+                for source in sources:
+                    if source not in self.JetUncertaintyBySources:
+                        raise RuntimeError("Jet energy scale uncertainty source requested, but not available")
+                    self.JetUncertaintyBySources[source].setJetEta(jet.eta())
+                    self.JetUncertaintyBySources[source].setJetPt(corr * jet.pt() * jet.rawFactor())
+                    unc = self.JetUncertaintyBySources[source].getUncertainty(True if delta>0. else False)
+                    group_shift += unc*unc
```
gpetruc: Does it ever happen that a specific uncertainty is supposed to have anti-correlated effects across different phase space regions?
In that case, squaring them up will break the feature.
If `len(sources) == 1`, when you don't need to do any addition, you could avoid the square and sqrt so that you would preserve any anti-correlation

gpetruc (changes requested):

GaelTouquet:
I have committed a fix for the requested change, but I haven't completely synced with other groups on the grouped systematics yet. It might not come from this code but from the analysis part, it is still under investigation. As said before, I will sync this ASAP and confirm here that it gives a satisfactory result.

PhysicsTools/Heppy/python/physicsutils/JetReCalibrator.py
GaelTouquet: Done

GaelTouquet:
Just a quick update : I have checked the code with an other group and everything seems to be alright but there is still disagreement on the synchronization of a test sample. Someone from the group agreed to do some scrutinized parallel testing, but he will only be available to do so beginning of next week.
Sorry to keep you waiting

GaelTouquet:
We have reached synchronization! The group I was syncing with had overwritten pointers...
Although we haven't decided if we should keep the ```len(sources) == 1``` case as was before, leading to the inversion of up and down shifts in this case.
I suggest to merge this the way it is and I will find a workaround in our analysis code if needed.
Thank you for your patience.

PR #756, opened 2019-06-22
Fixing FNAL condor script and removing obselete eid

bachtis:
There is an obselete 2012_VETO ID in autophobj that does not exist anymore in 10_X_Y
so I just commented it.
I also fixed the FNAL submission script
I will let someone else merge it


bachtis:
Should I merge this ?

gpetruc (approved):

